PR #6, opened 2020-03-11
testing pipeflow with unconnected junction

jkisse:
test to check if pipeflow converges with an unconnected junction in net.junctions


jkisse:
Travis CI not passing

PR #24, opened 2020-03-23
use pandas instead of numpy max function

lthurner:
In some combinations of numpy/pandas, it is not possible to apply np.max on a pandas Int64 Index, see:
https://github.com/numpy/numpy/issues/13285

I am experiencing this with numpy 1.17.4 and pandas 0.24.2. Using pandas .max() function instead should be more stable.


jkisse:
in the original PR, travis-CI did not respond, even though all tests passed. Closing an re-opending did not solve the issue. Merging an updated base branch seems to work as a workaround.

PR #62, opened 2020-04-27
adaptions in fluids.py

SimonRubenDrauz:
I made some changes in fluids:
- representation in a net/of a fluid
- preventing comp and der_comp from being hardcopied
- including functions to get molar mass, comp and der_comp
- for constants input variables can be given, but are not a must


SimonRubenDrauz:
@dlohmeier Could you have a look if the changes don't cause any problems. Mainly the part, where I said that a Property not necesserily needs an argument should be revised. Thanks!

SimonRubenDrauz:
I prevented the hardcopying of comp by defining a constant property which gets its data from the .txt-files. I prefer the .txt-files as they are more user-friendly. Espeically in case of the fluids and their properties a user might like to define their own fluid. A .json-file might be quite complex for some users and could be therefore deterrent. On top, what do you mean by too big. We already have the most import properties required for a pipeflow calculation and there are just 6 different properties per fluid. All fluids together some up to 56 kB. Therefore, I still prefer .txt or .csv-files over json-files.

SimonRubenDrauz:
About the tests: I totally agree with you. I could write one!

PR #85, opened 2020-06-10
update hydrogen properties

jkisse:
* hydrogen properties txt-files updated with data from NIST 
* hydrogen added to call_lib in fluids.py


SimonRubenDrauz:
Looks good to me. Could you also change the bugs in the other diatomic elements? Ohterwise I would do that!

jkisse:
Thanks Simon. I've already changed the properties for N2 and O2 on [this branch](https://github.com/jkisse/pandapipes/tree/correct_properties_values). However, this caused 8 failing tests which have to be modified.

Issue #86, opened 2020-06-16
Setting initial values for pipeflow


dlohmeier:
I would like to start a discussion on how the values of pressure, velocity and potentially temperature should be initialized in case of time series simulations. In pandapower, there are many different options available. Looking at the documentation, I find the following text:

    init (str, “auto”) - initialization method of the loadflow pandapower supports four methods for initializing the loadflow:
        - “auto” - init defaults to “dc” if calculate_voltage_angles is True or “flat” otherwise
        - “flat”- flat start with voltage of 1.0pu and angle of 0° at all PQ-buses and 0° for PV buses as initial solution
        - “dc” - initial DC loadflow before the AC loadflow. The results of the DC loadflow are used as initial solution for the AC loadflow.
        - “results” - voltage vector of last loadflow from net.res_bus is used as initial solution. This can be useful to accelerate convergence in iterative loadflows like time series calculations.

    init_vm_pu (string/float/array/Series, None) - Allows to define initialization specifically for voltage magnitudes. Only works with init == “auto”!
        - “auto”: all buses are initialized with the mean value of all voltage controlled elements in the grid
        - “flat” for flat start from 1.0
        - “results”: voltage magnitude vector is taken from result table
        - a float with which all voltage magnitudes are initialized
        - an iterable with a voltage magnitude value for each bus (length and order has to match with the buses in net.bus)
        - a pandas Series with a voltage magnitude value for each bus (indexes have to match the indexes in net.bus)

In general, I think that it makes sense to implement a similar strategy in pandapipes. However, we have a less rigid and in some places more implicit approach to building our internal structures. I see the following obstacles:
- Extracting results from results tables means that they have to be taken from all branch result tables for velocity and all node result tables for pressure and temperature. 
- If we consider internal nodes and edges, the results are not even included in any tables.

Therefore I think that it makes sense to only extract results from the internal structure (node_pit, branch_pit). This raises the following questions:
- Do we want to make sure that the previous pipeflow has led to valid results, or should this be left to the user?
[...]

PR #87, opened 2020-06-18
Speedup of timeseries calculation through additional option

dlohmeier:
I added the option reuse_internal_data. If it is set to True along with only_update_hydraulic_matrix, the jacobi csr-matrix and its lookup for node and edge equations are not deleted from the net after the pipeflow and will be reused in the next pipeflow. This drastically reduces the time spent in the function build_system_matrix.


SimonRubenDrauz:
To get it right: If you set the recycle flag in timeseries, this user options would be automatically added? Or how would you implement it differently in the timeseries module?

dlohmeier:
As of now, there is no recycle option implemented in pandapipes. In pandapower, this is done with the help of a dictionary that states which parts should be updated, so that the rest will just be recycled. I am taking an approach from the other side, i.e. for now I only want to state which part should be recycled. The rest of the pipeflow will be performed as usual. Either we will be able to add more and more options here and later summarize them in one kwarg, or maybe recycling will be addressed automatically with issue #86 . This is only a small first step.

SimonRubenDrauz:
Sounds good to me. I like your approach, as it step by step extends the speed-up options. Would it make sense to open a new issue or add it to #86 where we log all these options in order to have all options on the table when we want to implement a recycle in timeseries. By that we don't miss something. What do you think?

PR #102, opened 2020-07-29
Network created in master v0.1.0 for version checking added and tests created

ttrummel:


ttrummel:
release_control_test_network

Network for control between versions.

PR #104, opened 2020-07-29
Removed a STANET network from "networks" of OpenModelica

ttrummel:


SimonRubenDrauz:
Agree. Needs to be removed!

PR #107, opened 2020-08-03
Water Tower Component

SimonRubenDrauz:
Introducing new component: Water Tower


pandapipes/component_models/water_tower_component.py
```diff
@@ -0,0 +1,140 @@
...
+        :type node_pit:
+        :return: No Output.
+        """
+        water_towers = net[cls.table_name()]
+        density = net.fluid.get_density(water_towers.t_k.values)
+        junction_idx_lookups = get_lookup(net, "node", "index")[node_name]
+        junction = cls.get_connected_junction(net)
+        index_wt = junction_idx_lookups[junction]
+        press = density * (water_towers.height_m.values - node_pit[index_wt, HEIGHT]) * \
+                GRAVITATION_CONSTANT / P_CONVERSION * water_towers.in_service.values
```
jkisse, line 53: What exactly is the definition of height_m? My guess was the height of the water column in the tower. However, if the junction's height is subtracted, this could lead to negative height, couldn't it?
I'd find it confusing if height_m was in "meters above sea level" but I do not know the conventions.
As a suggestion, maybe it should be renamed to tower_height_m. A scaling factor could be introduced to derive water_column_height_m which then serves as input for the static pressure (rho*g*water_column_height_m)

pandapipes/component_models/water_tower_component.py
```diff
@@ -0,0 +1,140 @@
...
+    import pplog as logging
+except ImportError:
+    import logging
+
+logger = logging.getLogger(__name__)
+
+
+class WaterTower(NodeElementComponent):
+    """
+
```
jkisse, line 25: Could you add some information on the pictured model and important limitations?
It seems to me like an infinite diameter is assumed (no change in water column height on withdraw or feed-in), is that correct?
How does it differ from an ext_grid, apart from the different formulations of pressure, and to which extend does it overlap?

pandapipes/component_models/water_tower_component.py
SimonRubenDrauz: It actually should be the height above sea level. But I agree. It really seems confusing. It is a convention in SINCAL. But I would change it in our case taking only the height of the water_column.

pandapipes/component_models/water_tower_component.py
SimonRubenDrauz: Yes. I will do so.
There are not many differences compared to a ext_grid. The pressure is determined by the height of the water column. It is assumed that the water tower capacity is unlimited and the height of the water column remains constant, correct!
I agree that we could think of making a new parent component for ext_grid and water_tower. But for now I would leave it like this!

pandapipes/component_models/water_tower_component.py
```diff
@@ -0,0 +1,140 @@
...
+        :return: No Output.
+        """
+        water_towers = net[cls.table_name()]
+        density = net.fluid.get_density(water_towers.t_k.values)
+        junction_idx_lookups = get_lookup(net, "node", "index")[node_name]
+        junction = cls.get_connected_junction(net)
+        press = density * water_towers.height_m.values * \
+                GRAVITATION_CONSTANT / P_CONVERSION * water_towers.in_service.values
+        juncts_p, press_sum, number = _sum_by_group(junction.values, press,
+                                                    np.ones_like(press, dtype=np.int32))
```
jkisse, line 55: I am not sure whether this is physically possible. The [effect of communicating vessels](https://en.wikipedia.org/wiki/Communicating_vessels) (dt: ["kommunizierende Röhren"](https://de.wikipedia.org/wiki/Kommunizierende_R%C3%B6hren)) is not considered, is it?
I'd suggest limiting the number of water towers to 1 per net or raise a warning when towers with different water height are connected.

dlohmeier:
To me it is a bit questionable whether this component makes sense in this form, especially due to the similarity to the ext_grid that @jkisse mentioned. There are several issues connected to this component:
1. Why is it so important to create a whole component just for the sake of one equation that can easily be implemented by a user himself? If this is implemented as a controller with an ext_grid, it is possible to adapt the pressure of the ext_grid with the help of the water height, and you could even implement a control loop that adapts the average height during a time series simulation based on the water extraction.
2. If you implement this as a component, what do you do if an ext_grid is connected to the same junction as a water tower and they don't show the same pressure? What you implemented by now only divides the mass flows over those components in the results, but the pressure PINIT can only be set in each component individually, so the latest initialized component determines the pressure.
Maybe this is the time where we should think about how to aggregate different components and their behavior in one function, which would also lead to a speed-up if e.g. all pipes and valves can be calculated as one part of the pit. Yet, this adds more complexity to our whole structure which is already extremely complex.

pandapipes/component_models/water_tower_component.py
jkisse: Thanks for the adaption! However, thinking about it twice, maybe this leads to a problem regarding communicating vessels. (see my latest comment.) This could be the reason why Sincal uses absolute height.

SimonRubenDrauz:
Thansk for your feedback. I adapted now everything. What do you think?

PR #108, opened 2020-08-05
Adaption ppipe hook

SimonRubenDrauz:
adaption of ppipe_hook: increasing recycability and enabling compatibility with pandapower

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
   - if [ "$TRAVIS_BRANCH" = "master" ]; then
          git clone --single-branch --branch master https://github.com/e2nIEE/pandapower.git --depth 1 /tmp/pandapower;
     else
-         git clone --single-branch --branch develop https://github.com/e2nIEE/pandapower.git --depth 1 /tmp/pandapower;
+         git clone --single-branch --branch adaption_pp_hook https://github.com/SimonRubenDrauz/pandapower.git --depth 1 /tmp/pandapower;
     fi
   - pip install /tmp/pandapower
   - conda install matplotlib --yes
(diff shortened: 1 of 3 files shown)


dlohmeier:
I would like to try and use as little changes between pandapipes and pandapower as possible. In general, there is only one difference between pp_hook and ppipes_hook, which is the registry that is not even called in that very function. So I would like to suggest the following changes:

- in pandapower: remove pp_hook_serialization

        class PPJSONDecoder(json.JSONDecoder):
            def __init__(self, **kwargs):
                # net = pandapowerNet.__new__(pandapowerNet)
                net = create_empty_network()
                super_kwargs = {"object_hook": partial(pp_hook, net=net)}
                super_kwargs.update(kwargs)
                super().__init__(**super_kwargs)


        def pp_hook(d, net=None, registry_class=FromSerializableRegistry):
            if '_module' in d and '_class' in d:
                if "_object" in d:
                    obj = d.pop('_object')
                elif "_state" in d:
                    obj = d['_state']
                    if d['has_net']:
                        obj['net'] = 'net'
                    if '_init' in obj:
                        del obj['_init']
                    return obj  # backwards compatibility
                else:
                    # obj = {"_init": d, "_state": dict()}  # backwards compatibility
                    obj = {key: val for key, val in d.items() if key not in ['_module', '_class']}
                class_name = d.pop('_class')
                module_name = d.pop('_module')
                fs = registry_class(obj, d, net, pp_hook)
                fs.class_name = class_name
                fs.module_name = module_name
                return fs.from_serializable()
            else:
                return d

- in pandapipes: remove ppipes_hook and ppipes_hook_serialization and FromSerializableRegistryPpipe.__init__() (super-init is called automatically)

        class PPipeJSONDecoder(PPJSONDecoder):
            def __init__(self, **kwargs):
                obj_hook= kwargs.pop("object_hook", partial(pp_hook, net=net, registry=FromSerializableRegistryPpipe))
                super().__init__(object_hook=obj_hook, **kwargs)

Then we just have to change a few imports and use the PPipeJSONDecoder in the pandapipes file_io. Another option would be to just use object_hook = partial(pp_hook, net=net, registry=FromSerializableRegistryPpipe) in from_json directly, then we don't even need the PPipeJSONDecoder.

PR #113, opened 2020-08-12
Correct properties values for bi-atomic gases

jkisse:
and update respective tests


jkisse:
I'll force-merge, because it's just some data changes and all tests pass.

PR #119, opened 2020-08-14
adaptions in create-docstring

jkisse:
For some reason, some create functions had no line breaks in the list of parameters in the compiled documentation.

These changes should fix this.


jkisse:
It did not work. In RTD, there are still not all line breaks in the parameter documentation of create_pipe, create_heat_exchanger, and create_circ_pump. I don't understand why — probably due to different package versions.

jkisse:
New theory: This might be related to very long example-boxes. There are no line breaks in the examples.

PR #120, opened 2020-08-18
Split junction function

jkisse:
A toolbox function that can be helpful if a new branch element should be inserted into an existing network. It creates a new junction on the same position as the old one that has to be provided; a little offset can be added for differentiation in plots. Then, the from_junctions or to_junctions of all branch elements that were connected to the old junction are changed to the new junction. Node elements remain with the old junction. 

In a second step (not part of this function), a branch component (e.g., pump) has to be inserted between the old junction and the new junction. Otherwise, the junctions are not connected. 

What do you think about? Also, if a better name comes to your mind, I am happy to change it :)

### DIFF
--- a/pandapipes/test/test_toolbox.py
+++ b/pandapipes/test/test_toolbox.py
@@ -201,6 +201,39 @@ def test_create_continuous_index(create_net_changed_indices):
     for comp in previous_junctions.keys():
         assert np.all(net[comp].index == np.arange(len(net[comp])))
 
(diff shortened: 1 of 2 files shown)


jkisse:
As a start, this should not become a generic function but a special, specific (private) function in the required module

jkisse:
PR will be closed and can be re-opened when there is a precise need for this function

jkisse:
PR will be closed and can be re-opened when there is a precise need for this function

PR #122, opened 2020-08-19
convert format

SimonRubenDrauz:
This proves if the grids of different versions are still convertable.

There is a bug in pandapower. As soon as it is fixed and merged, the test should pass.


pandapipes/io/convert_format.py
```diff
@@ -0,0 +1,44 @@
...
+            else:
+                net['controller'].drop('controller', inplace=True, axis=1)
+        net["controller"].rename(columns={"controller": "object"}, inplace=True)
+
+
+def _update_initial_run(net):
+    if "controller" in net:
+        for ctrl in net.controller.object.values:
+            if hasattr(ctrl, 'initial_pipeflow'):
+                ctrl.initial_run = ctrl.initial_pipeflow
```
dlohmeier, line 38: Should initial_pipeflow be removed?

pandapipes/io/convert_format.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def reset_results(net):
```
dlohmeier, line 41: Does that mean that results will always be removed in old versions? I think that is not what was intended in pandapower. We want to keep results inside the tables as much as possible.

pandapipes/io/file_io.py
```diff
@@ -120,10 +121,10 @@ def from_json(filename):
     else:
         with open(filename) as fp:
             json_string = fp.read()
-    return from_json_string(json_string)
+    return from_json_string(json_string, convert=convert)
 
 
-def from_json_string(json_string):
+def from_json_string(json_string, convert=False):
```
dlohmeier, line 127: Why is the default True in from_json and False in from_json_string?

pandapipes/test/api/test_convert_format.py
```diff
@@ -0,0 +1,45 @@
...
+    v_mean_m_per_s_old = net.res_pipe.v_mean_m_per_s.copy()
+    convert_format(net)
+    try:
+        pp.pipeflow(net, run_control="controller" in net and len(net.controller) > 0)
+    except:
+        raise UserWarning("Can not run pipe flow in network "
+                          "saved with pandapipes version %s" % version)
+    p_res = np.allclose(p_bar_old.values, net.res_junction.p_bar.values)
+    v_res = np.allclose(v_mean_m_per_s_old.values, net.res_pipe.v_mean_m_per_s.values)
+    if not all([p_res, v_res]):
```
dlohmeier, line 39: Why isn't this an assert statement?

dlohmeier (commented):
Good to have that feature. I added some comments to aspects that I think need to be addressed still.

pandapipes/io/file_io.py
SimonRubenDrauz: This is how pandapower did it. Therefore, I did it the same way.

pandapipes/test/api/test_convert_format.py
SimonRubenDrauz: We could do that, however, I also sticked to the pandapower test_convert_format

pandapipes/io/convert_format.py
SimonRubenDrauz: Yes, right, pandapower is doing the same in my opinion. When I load for example:
net = nw.mv_oberrhein() and run reset_results from pandapower, I delete all results.
I think that it should make sure that the res_tables are always comprise the correct columns. As in our cases the res_tables are created differently, we could think about keeping the results.

dlohmeier: Actually my comment refers rather to the convert_format function. There you state that if the version in net is smaller than the current pandapipes version, the convert-functions will be called. In that case, reset_results will be called for all older versions, although there is no conversion problem.

pandapipes/io/convert_format.py
SimonRubenDrauz: I also didn't remove it in pandapower. Florian Schäfer recommended not to do this right after the first release, but wait a bit longer. Maybe I could add a warning like logger.warning("initial_pipeflow is deprecated, but it is still an attribute in your controllers. It will be removed in the future. Please use initial_run instead!")

SimonRubenDrauz:
Thanks for the comments. Find my answers above.

pandapipes/io/convert_format.py
SimonRubenDrauz: I see what you mean. I would remove it then!

dlohmeier (approved):

PR #124, opened 2020-08-26
add H2 as one default fluid in doc and tutorials, fall back option in add_fluid

jkisse:
* Hydrogen is now an additional default fluid and has been added to the text in the tutorials and documentation.
* if someone passed a string to add_fluid_to_net(net, fluid_as_string), it would write the string to the net. A call_lib intermediate step has been added.
* import of run_timeseries in the time series tutorial fixed


dlohmeier (changes requested):
Please add a logging info for add_to_net if string is passed.

dlohmeier (approved):

PR #133, opened 2020-09-04
adapted io operations to match new pandapower structure; this include…

dlohmeier:
…s saving addresses for objects and only references if they are saved anywhere else; small adaptations to classes inheriting from JSONSerializableClass


dlohmeier:
This should be merged only when the pandapower structure has changed (until then the tests will fail (cf. pandapower PR 909))

dlohmeier:
Closed because pandapower restructuring was not implemented. If it is, please reopen.

PR #141, opened 2020-09-16
Clean control

SimonRubenDrauz:
removing net to prevent circular references
(only merge after clean_control_2.0 has been merged in pandapower)

### DIFF
--- a/pandapipes/io/io_utils.py
+++ b/pandapipes/io/io_utils.py
@@ -8,6 +8,7 @@ from functools import partial
 from inspect import isclass
 
 import pandapower as pp
+import pandapipes as ppipes
 from pandapipes.component_models.abstract_models import Component
 from pandapipes.create import create_empty_network as create_fluid_network
 from pandapipes.pandapipes_net import pandapipesNet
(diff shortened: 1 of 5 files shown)


pandapipes/io/io_utils.py
```diff
@@ -62,8 +51,9 @@ def pandapipesNet(self):
             from pandapipes import from_json_string
             return from_json_string(self.obj)
         else:
-            self.net.update(self.obj)
-            return self.net
+            net = create_empty_network()
```
dlohmeier, line 54: I think this part is still somewhat unnecessary, but we could leave it inside for now. We should check, if pandapipesNet.__new__() can be called and the net.update(self.obj) leads to the same results. Missing tables and columns from earlier versions should be added in the convert_format script.

dlohmeier (approved):
I like the changes. We should check if we can still improve the performance (see my comment), but otherwise this looks very clean.

pandapipes/io/io_utils.py
SimonRubenDrauz: Cool, never thought about that. We definitely should check that.

What do you mean by convert_format comment. What tables and columns do you have in mind, can't think of any right now.

dlohmeier: If we add a new table in future (e.g. "compressor"), it will not appear in old versions. But instead of creating a new network to have all the tables created and then just overwriting all tables except compressor, we should just check for missing tables in the `convert_format` script and add them there.

SimonRubenDrauz: I see, yes, we should keep this in mind. I will open a new issue.

PR #144, opened 2020-09-24
bugfix in connectivity check

SimonRubenDrauz:
- bugfix connective check
- new test to cover this bug


pandapipes/pipeflow_setup.py
```diff
@@ -602,11 +603,6 @@ def reduce_pit(net, node_pit, branch_pit, nodes_connected, branches_connected):
         net["_lookups"]["branch_index_active"] = copy.deepcopy(get_lookup(net, "branch", "index"))
     else:
         active_pit["branch"] = np.copy(branch_pit[branches_connected, :])
-        if reduced_node_lookup is not None:
```
dlohmeier, line 605: I am wondering if we need the lookup from line 600 here as well...

dlohmeier: The lookups for branches within the active pit, I mean...

PR #147, opened 2020-10-13
changes in nxgraph creation to be based on component models

dlohmeier:
- usage of component models (Pipe, Valve, Pump) to generate nxgraph edges in very generalized way
- added identifiers to/from node columns to components
- added possibility to neglegt active identifier of specified branch components (mainly used for valves)


pandapipes/test/topology/test_nxgraph.py
```diff
@@ -19,8 +20,18 @@ def test_include_branches():
 
     mg = top.create_nxgraph(net, include_pipes=True, include_valves=True, include_pumps=False)
     assert len(mg.nodes()) == len(net.junction)
-    assert len(mg.edges()) == len(net.pipe) + len(net.valve) 
+    assert len(mg.edges()) == len(net.pipe)
```
SimonRubenDrauz, line 23: I don't get, why the valves are not taken into account, even though "include_valves" is set to True.

pandapipes/test/topology/test_nxgraph.py
```diff
@@ -19,8 +20,18 @@ def test_include_branches():
...
+
+    mg = top.create_nxgraph(net, include_pipes=True, include_valves=True, include_pumps=False,
+                            neglect_in_service=["valve"])
```
SimonRubenDrauz, line 26: Instead of neglect_in_service, I would rather do it like in plotting: respect_in_service. That would mean for me, if I set include_valves=True, then I would say all edges are considered, no matter if they are in_service or not. However, when to say, respect_in_servce=['valve'], I would  only take the valves which are in_service. As in this particular example, both valves are in_service, therefore, I would say, in both cases len(mg.edges()) == len(net.pipe) + len(net.valve) should be correct. If you set both out-of-service and set respect_in_service=['valves'], assert len(mg.edges()) == len(net.pipe)  should be correct.

pandapipes/topology/create_graph.py
```diff
@@ -22,38 +25,24 @@
...
-            add_edges(mg, indices, parameter, in_service, net, "pump")
+    if neglect_in_service is False:
+        neglect_in_service_comps = []
+    elif neglect_in_service is True:
+        neglect_in_service_comps = [comp.table_name() for comp in net.component_list]
+    else:
+        neglect_in_service_comps = list(neglect_in_service)
+
+    for comp, weight_col, include_comp in zip([Pipe, Valve, Pump], ["length_km", None, None],
+                                              [include_pipes, include_valves, include_pumps]):
```
SimonRubenDrauz, line 43: What if someone creates another, new branch_element and wants to add this to the nx_graph. This, as I can see it, is not possible at the moment. How about adding a **kwargs, where you can pass a list of additional components you want to add.

pandapipes/topology/create_graph.py
```diff
@@ -22,38 +25,24 @@
...
+        if comp in net.component_list:
+            add_branch_component(comp, mg, net, include_comp, neglect_in_service_comps, weight_col)
```
SimonRubenDrauz, line 45: I like the much cleaner version. Really nice idea!

SimonRubenDrauz (changes requested):

pandapipes/test/topology/test_graph_searches.py
```diff
@@ -6,13 +6,20 @@
...
 
+
 def test_connected_components():
     net = nw.gas_meshed_two_valves()
     mg = top.create_nxgraph(net)
     assert len(list(top.connected_components(mg))) == 1
     mg = top.create_nxgraph(net, include_valves=False)
     assert len(list(top.connected_components(mg))) == 2
     mg = top.create_nxgraph(net, include_pipes=False)
+    assert len(list(top.connected_components(mg))) == 7
```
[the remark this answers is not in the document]
dlohmeier: The reason here is simply that in the previous version valves were never considered wrt opened/closed, so all opened and closed valves were included in the nxgraph. It was simply undesired behavior.

pandapipes/test/topology/test_nxgraph.py
dlohmeier: I don't really agree with that point for two reasons:
1. It would alter the behavior from pandapower, making it harder for users of pandapower to use the functionality. In pandapower, in_service is always respected and branches are excluded in that case.
2. To me, the pandapower beahvior is more straight-forward. However, in pandapower, neglecting the in_service flag is only possible for a hand full of elements and it's hard coded.
Maybe we can combine it with your next comment and somehow add keyword arguments that can respect the behavior for certain components...

pandapipes/topology/create_graph.py
dlohmeier: Yes, I thought about that as well... but I haven't yet found a way to insert the weight behavior here. I am sure, we can come up with a nice solution, especially since we need to combine it with your previous comment.

pandapipes/test/topology/test_nxgraph.py
dlohmeier: See my statement above. I may write:
assert len(mg.edges()) == len(net.pipe) + net.valve.in_service.sum()

pandapipes/component_models/abstract_models/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .component_models import *
+from .base_component import *
```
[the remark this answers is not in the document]
dlohmeier: OK...

pandapipes/topology/create_graph.py
```diff
@@ -75,8 +64,38 @@ def create_nxgraph(net, include_pipes=True, include_valves=True, include_pumps=T
...
+    indices = np.zeros((n, 3), dtype=np.int)
+    indices[:, INDEX] = tab.index
+    parameters = np.zeros((n, 1), dtype=np.float)
+
+    if neglect_in_service:
+        return indices, parameters, np.ones(n, dtype=np.bool)
+    elif in_service_col in tab:
+        return indices, parameters, tab[in_service_col].values.copy()
+    else:
+        return indices, parameters
```
SimonRubenDrauz, line 101: Isn't it possible to get a more generalized version both useable for pandapower and pandapipes

pandapipes/component_models/abstract_models/branch_models.py
```diff
@@ -28,6 +28,22 @@
...
+    def table_name(cls):
+        return NotImplementedError
+
+    @classmethod
+    def get_component_input(cls):
+        return NotImplementedError
+
+    @classmethod
+    def get_result_table(cls, net):
+        return NotImplementedError
```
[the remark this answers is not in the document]
dlohmeier: I will create another PR, agreed. I find it a little cleaner to define not implemented inherited methods to clarify for everyone inheriting from that class that it is abstract and that all abstract methods should be defined in the inheriting class.

pandapipes/component_models/abstract_models/branch_models.py
```diff
@@ -28,6 +28,22 @@
...
+
+    @classmethod
+    def from_to_node_cols(cls):
+        raise NotImplementedError
+
```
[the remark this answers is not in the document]
dlohmeier: Yes, this is very important, as we are using generic functions, where the column names are not defined, but inquired from the component model. I.e. you might define your own branch component that uses "junction1" and "junction2" instead of "from_junction" and "to_junction".

pandapipes/test/topology/test_nxgraph.py
SimonRubenDrauz: Okay, I see. I actually don't really like, that plotting and topology is the exact opposite. In plotting you can set respect_in_service, because the default option is that everything is considered. In topology, however, default is to only consider in_service and then neglect only in_service if the flag is set. I think, we should be more consistent. However, if it is done like this in pandapower, then we should stick to that. Might be a aspect that can be considered in pandapower then.

pandapipes/topology/create_graph.py
dlohmeier: Probably... however, that requires a few changes in pandapower as well... we can see, if we can make them under the hood.

pandapipes/component_models/abstract_models/branch_models.py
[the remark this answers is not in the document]
SimonRubenDrauz: Okay, I wasn't that deep in that topic that I understood everthing in detail. Thanks for the clarification :). We should keep it here then.

pandapipes/topology/create_graph.py
SimonRubenDrauz: Okay, maybe add a todo and add a issue to pandapower and pandapipes. Till then, I would merge the PR.

pandapipes/topology/create_graph.py
dlohmeier: OK, here's a suggestion: We always add all branch components and respect the in_service flag. Only if clearly indicated otherwise in the kwargs, we alter the behavior:

    for comp in net.component_list:
        if not issubclass(comp, BranchComponent):
            continue
	tbl = comp.table_name()
	include_comp = kwargs.get("consider_%s" % tbl, True)
	respect_in_service = kwargs.get("respect_%s" % tbl, True)
	# some formulation to add weight
	weight_col = ??
	add_branch_component(comp, mg, net, include_comp, respect_in_service, weight_col)

And then we just need to rewrite the `add_branch_component` function.

SimonRubenDrauz: Maybe don't use it in kwargs, but rather do it like included_components=None and weight='length_km'. Alternatively, there you can specify included_components and weight indivudally. included_components can be a list then and the weight-variable must be a list of same length, too. If included_components is None, all branch_components are taken into account and 'legnth_km' is taken as default. What do you think or do you think it is too implicit?

dlohmeier:
I added other PRs to do the changes more step-by-step. Hope we can come up with a clear structure here. That's why I added the donotmerge label.

dlohmeier:
This was now implemented in PR #178

PR #159, opened 2020-11-05
bugfix in pipeflow when there are only one node nets 

SimonRubenDrauz:
Solving problem if there is only a one node net without branches. Corresponding tests added!


jkisse (approved):
great, looks like a very precise and elegant solution 😃🏆

dcronbach (approved):
Ich denke, das sollte so gehen.

PR #166, opened 2020-11-26
update Pump std type in order to still be able to access the original pressure and flowrate data if they exist

SimonRubenDrauz:
In pandagrids we need the original pressure and flowrate values, therefore, I need to keep them in the Pump class.

### DIFF
--- a/pandapipes/std_types/std_type.py
+++ b/pandapipes/std_types/std_type.py
@@ -54,7 +54,7 @@ class PumpStdType(StdType):
 
     def update_pump(self, pressure_list, flowrate_list, reg_polynomial_degree):
         reg_par = regression_function(pressure_list, flowrate_list, reg_polynomial_degree)
-        self.res_par = reg_par
+        self.reg_par = reg_par
         self._pressure_list = pressure_list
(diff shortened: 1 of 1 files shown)


pandapipes/std_types/std_type.py
```diff
@@ -48,6 +48,17 @@ def __init__(self, name, reg_par):
         """
         super(PumpStdType, self).__init__(name, 'pump')
         self.reg_par = reg_par
+        self._pressure_list = None
+        self._flowrate_list = None
+        self._reg_polynomial_degree = 2
+
+    def update_pump(self, pressure_list, flowrate_list, reg_polynomial_degree):
+        reg_par = regression_function(pressure_list, flowrate_list, reg_polynomial_degree)
+        self.res_par = reg_par
```
jkisse, line 57: do you really mean self.re**s**_par here?

SimonRubenDrauz: you are right. That is a type. Should be self.reg_par.

jkisse (approved):
looks good to me

PR #172, opened 2020-12-01
usage of new for pandapipesNet in from_json

dlohmeier:


dlohmeier:
This is now done in PR #223 .

PR #175, opened 2020-12-03
add GitHub Actions CI for develop and master

jkisse:
GitHub Actions is free for public repos and fast. Maybe a good alternative to Travis CI. What do you think, shall we switch over?
There is a similar PR in pandapower https://github.com/e2nIEE/pandapower/pull/1031


dlohmeier:
Are you sure that the CI is for free? I thought that there is a limit, just like with Travis CI. Maybe we should exchange some links for further reading. At https://github.com/pricing it says: 2000 Actions minutes/month; Free for public repositories. Does it mean that the limit only applies to private repositories?

jkisse:
I think GitHub Actions for public repositories are not getting billed... I did quite some tests yesterday and today and it still says "0/2000 min used"

dlohmeier (commented):
Would it be possible to integrate master and develop version in one file? Then there could be a flag in the pandapower installation command specifying the respective branch. The same might be possible for lint (if not required on master branch).

jkisse:
It is possible to add conditionals for branches like this
`if: github.ref == 'refs/heads/master'`

However, I am not sure how requests that want to merge into the master branch can be covered — because the triggering branch would still be "develop", wouldn't it?

PR #196, opened 2021-01-07
Cleanup

SimonRubenDrauz:
cleaning up source code

### DIFF
--- a/pandapipes/component_models/abstract_models/branch_models.py
+++ b/pandapipes/component_models/abstract_models/branch_models.py
@@ -14,8 +14,8 @@ from pandapipes.idx_branch import FROM_NODE, TO_NODE, LENGTH, D, TINIT, AREA, K,
     JAC_DERIV_DV_NODE, \
     LOAD_VEC_BRANCHES, LOAD_VEC_BRANCHES_T, LOAD_VEC_NODES_T, ELEMENT_IDX
 from pandapipes.idx_node import PINIT, HEIGHT, TINIT as TINIT_NODE, PAMB
-from pandapipes.internals_toolbox import _sum_by_group
-from pandapipes.pipeflow_setup import get_table_number, get_lookup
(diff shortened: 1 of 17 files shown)


pandapipes/component_models/abstract_models/branch_models.py
```diff
@@ -265,33 +265,111 @@ def calculate_pressure_lift(cls, net, pipe_pit, node_pit):
...
+        _, active_pipes = _sum_by_group(idx_pit, pipe_considered.astype(np.int32))
         active_pipes = active_pipes > 0.99
         placement_table = placement_table[active_pipes]
         branch_pit = net["_active_pit"]["branch"][fa:ta, :]
+        return placement_table, branch_pit, res_table
+
 
-        return placement_table, branch_pit, results
+    @classmethod
+    def extract_results(cls, net, options, node_name):
```
dlohmeier, line 301: Was this moved from the pipe component? Why can this be generalized?

SimonRubenDrauz: Yes, this move was from pipe component (and also more or less from heat_exchanger and valve, as they have the same extract results function). The main reason why I did this was the following: I realized that in pipe component there were changes in extract_results which I couldn't find in valve and heat exchange component (duplicated code problem -.-). Therefore I decided to bring it up so that all of them inherit the same function (their common class is the branch_component). I agree that it seems to be not generalizable, but I bringt up the following argument: I say that from all branch_components these extract results are the most general. If not, you need to adapt it as user.

Alternatively, in order to prevent duplicated code I was thinking about to move it in a toolbox function which all of them can just use. But I prefer this solution more as it keeps the most relevant parts all at one place.

dlohmeier (approved):
I think it looks all fine. However, I am not sure if the generalization of result extraction on branch level might lead to problems in future. There should be a discussion on how to generalize some functions. If we can do the same for many components, we should only do it once.

PR #202, opened 2021-01-08
license update 2021

dlohmeier:
Just updated the copyright in all the license texts.


SimonRubenDrauz:
I did the same already :D and added also University of Kassel.

dlohmeier:
Hmm, that is strange, I didn't see that. Then I will close the PR.

PR #215, opened 2021-01-19
Bugfix in control and time series

SimonRubenDrauz:
bugfix in control and timeseries due to changes in pandapower

### DIFF
--- a/pandapipes/control/run_control.py
+++ b/pandapipes/control/run_control.py
@@ -35,18 +35,9 @@ def prepare_run_ctrl(net, ctrl_variables, **kwargs):
     :return: ctrl_variables
     :rtype: dict
     """
-    ctrl_var = ctrl_variables
-
     if ctrl_variables is None:
-        ctrl_variables  = prepare_run_control_pandapower(net, None)
+        ctrl_variables  = prepare_run_control_pandapower(net, None, **kwargs)
         ctrl_variables["run"] = ppipe.pipeflow
-
-    if ('continue_on_divergence') in kwargs and (ctrl_var is None):
-        div = kwargs.pop('continue_on_divergence')
-        ctrl_variables['continue_on_divergence'] = div
-    if ('check_each_level') in kwargs and (ctrl_var is None):
-        check = kwargs.pop('check_each_level')
-        ctrl_variables['check_each_level'] = check
 
     ctrl_variables["errors"] = (PipeflowNotConverged,) # has to be a tuple
 
--- a/pandapipes/control/run_control.py
+++ b/pandapipes/control/run_control.py
@@ -21,12 +21,12 @@ def run_control(net, ctrl_variables=None, max_iter=30, **kwargs):
     :return: No output
     """
     if ctrl_variables is None:
-        ctrl_variables = prepare_run_ctrl(net, None)
+        ctrl_variables = prepare_run_ctrl(net, None, **kwargs)
 
     run_control_pandapower(net, ctrl_variables=ctrl_variables, max_iter=max_iter, **kwargs)
 
 
-def prepare_run_ctrl(net, ctrl_variables):
+def prepare_run_ctrl(net, ctrl_variables, **kwargs):
     """
     Function that defines default control variables.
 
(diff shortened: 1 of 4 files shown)


pandapipes/control/run_control.py
```diff
@@ -35,10 +35,19 @@ def prepare_run_ctrl(net, ctrl_variables):
     :return: ctrl_variables
     :rtype: dict
     """
+    ctrl_var = ctrl_variables
+
     if ctrl_variables is None:
         ctrl_variables  = prepare_run_control_pandapower(net, None)
         ctrl_variables["run"] = ppipe.pipeflow
 
+    if ('continue_on_divergence') in kwargs and (ctrl_var is None):
```
jkisse, line 44: Why is this `and (ctrl_var is None)`? I thought a case with some ctrl_var (but not continue_on_divergence) and continue_on_divergence in the kwargs could also happen, couldn't it?
I'd expect something like `and ((ctrl_var is None) or (continue_on_divergence not in ctrl_var.keys()))`

pandapipes/control/run_control.py
```diff
@@ -35,10 +35,19 @@ def prepare_run_ctrl(net, ctrl_variables):
...
+        div = kwargs.pop('continue_on_divergence')
+        ctrl_variables['continue_on_divergence'] = div
+    if ('check_each_level') in kwargs and (ctrl_var is None):
```
jkisse, line 47: If the if in L44 was True, this will always be False because ctrl_var is not None.
Like above, I'd expect something like `and ((ctrl_var is None) or (continue_on_divergence not in ctrl_var.keys()))`

pandapipes/multinet/control/run_control_multinet.py
```diff
@@ -260,6 +267,16 @@ def prepare_run_ctrl(multinet, ctrl_variables):
         ctrl_variables[net_name]['only_v_results'], ctrl_variables[net_name]['recycle'] = \
             get_recycle(ctrl_variables_net)
 
+        if ('continue_on_divergence') in kwargs and (ctrl_var is None):
```
jkisse, line 270: Same as in the comment above. I'd expect something like `and ((ctrl_var is None) or (continue_on_divergence not in ctrl_var.keys()))`

pandapipes/multinet/control/run_control_multinet.py
```diff
@@ -239,13 +240,19 @@ def prepare_run_ctrl(multinet, ctrl_variables):
                     ctrl_variables[net_name]['initial_run'] is not None else ctrl_variables_net['initial_run']
                 ctrl_variables[net_name]['only_v_results'], ctrl_variables[net_name]['recycle'] = \
                     get_recycle(ctrl_variables_net)
+
+                if ('continue_on_divergence') in kwargs and (ctrl_var is None):
```
jkisse, line 244: Same as in the comment above. I'd expect something like `and ((ctrl_var is None) or (continue_on_divergence not in ctrl_var.keys()))`

pandapipes/control/run_control.py
SimonRubenDrauz: Following cases I was thinking about:
- a user defines ctrl_var and continue_on_divergence, then always the ctrl_var is preferred.
- Only if ctrl_var is None, then look for a continue_on_divergence in kwargs
- in the default prepare_run_control_pandapower by default continue_on_divergence is set to False
Actually, what I see now is that you don't need this lines but you can just pass the kwargs. So I would change that if you agree.

In your case you say that the user doesn't need to define continue_on_divergence. I wouldn't do that as, if the user wants to define his/her own ctrl_variables, he should be aware of what it is doing.

SimonRubenDrauz: Maybe something like:
if ('continue_on_divergence') in kwargs and (ctrl_var is None or not 'continue_on_divergence'  in ctrl_var.keys())
following use cases: 
1. user defines ctrl_variables and no divergence, it takes divergence from ctrl_variables (user needs to define it)
2. user defines ctrl_variables (without divergence) and divergence, divergence is taking from the kwargs
3. user defines both, ctrl_varibales is preferred
4. user defines no ctrl_variables but divergence, divergence is used.

Same solution could be selected in case of check_each_level. What do you think?

jkisse: The suggestion from the last post sounds good to me. I am not sure though, whether in Case 3 (user defines both) ctrl_variables should be preferred, as it is slightly more indirect.
In the end, I think the main point is simply to have the behavior documented somewhere.

SimonRubenDrauz: Actually I think if a user takes the effort and defines ctrl_variables himself/herself, I don't think it is implicit especially if you think about the following case: In a multinet you can define ctrl_variables for each net. So you define divergence for each net and think about what happens if powerflow is not converging. 
So what would happen if gloabl strikes net specific: the globally defined divergence would overwrite all the ctrl_variables net specific divergence entries. I implemented it that way that it would only change divergence if there is none. If it is defined by the user for a net specifically, it won't change the entry. Therefore, I would like stick to this solution. 
But we could open a new issue if you like and discuss it in a bigger plenum.

jkisse: Okay, I see your point. Sounds good to me.

Issue #142, opened 2020-09-17
IO-Utils 


SimonRubenDrauz:
check, if pandapipesNet.__new__() can be called and the net.update(self.obj) leads to the same results, instead of creating an empty network in io-utils. Missing tables and columns from earlier versions should be added in the convert_format script, instead of just overwriting everthing in io-utils.

PR #223, opened 2021-01-26
Suggestion for default components

dlohmeier:
in io_utils, a pandapipes net is initialized almost empty; default components added separately and also added to convert_format


dlohmeier:
A solution for issue #142.

Issue #211, opened 2021-01-13
Maximum iterations of Colebrook-White algorithm


dlohmeier:
I am currently running some calculations of different network topologies for a large gas network and found that the Colebrook-White algorithm does not use a limit for its iterations. However, in some cases the algorithm obviously does not converge, as I had to kill my script after nearly half a million iterations. I think it would be wise to implement a maximum number of iterations and use a normal pipeflow exception, if this number is reached. But endless loops should really be avoided.

@dcronbach do you have a suggestion on a suitable default number of maximum iterations?

dcronbach:
If I remember correctly, the algorithm typically converges in only very few iterations. If you set it to 100, it might be more than enough. I would use this as a start and see, if this number has to be adapted in the future. 

Nevertheless, I do not know if this number is feasible, as I have not done larger calculations in pandapipes since ~1 year.

dlohmeier:
Yes, I think that it should usually converge very fast. If it does not, that is probably caused by erroneous input data and accompanied by problems with the NR algorithm. So, if using the Nikuradse approach, the NR algorithm wouldn't converge either, so that throwing a ppException would totally make sense.

PR #225, opened 2021-01-28
pipeflow exception is raised if colebrook algorithm does not converge

dlohmeier:


dlohmeier:
This should solve Issue #211.

dcronbach (approved):
Looks good. The import statement in derivative_toolbox in line 48 looks a little bit odd as we usually put them on top of each module. But you probably placed it there because the import only happens because of the exception.

dlohmeier:
Yes, I would prefer a top-level import as well, but that would lead to a circular import that cannot be resolved. I think there should be another way to handle this problem, but I am not yet sure how. I mentioned this problem in issue #232.

dlohmeier:
I added a TODO statement there.

Issue #199, opened 2021-01-07
generalizing code


SimonRubenDrauz:
To generalize calculate derivatives and extract results on branch_models.py level might lead to problems in near future, as a most abstract class defines really real and concrete functions/classmethods. Therefore, there should be a discussion on how to make the code as abstract as possible, prevent duplicated code and seperate it from really concrete functions which should only be part of the real component models.

dlohmeier:
This problem should be tackled with the numba development (cf. dev_numba branch or PR #335 ). What I suggest there is:
- make calculate_derivatives a global function to be called on the whole branch pit (actually, almost all components do nearly exactly the same thing, as you pointed out)
- implement two class methods for all components for a priori and a posteriori adaption of the derivative values
Of course this means that we have to take care of a very general behavior for all types of branches, but we actually do that already and still go through all components individually instead of performing this step just once.

dlohmeier:
The dev_numba branch is merged. The only feature we still miss somehow is a possibility to tell the global result extraction function which results to extract. Currently, this is somewhat hard coded and the components can decide whether they need these results and could extract further results for their own, but maybe there is a smarter way to do this.

PR #226, opened 2021-01-28
Dev numba

dlohmeier:
I would like to suggest some changes to the pipeflow procedure:

- All derivatives (for now only implemented in hydraulic calculations) are calculated globally, i.e. the branch component's calculate_derivatives_hydraulic is moved outside that class, as this behaviour is always the same.
- It is possible to define your own additional procedures for updating the pit before and after derivative calculation. This is helpful for instance in case of pressure lift recalculation or for branch components whose pressure difference does not depend on the flow.
- I implemented some parts in numba which should accelerate the pipeflow calculation in most cases.

Numba does have downsides that should be discussed:
- It is very hard to debug numba code.
- Compilation takes time, which leads to the fact that in case of just one pipeflow for a small network, this solution is even slower. This can be seen from the test timing, which will be slower with the numba implementations. The larger the dataset and the more times the function is called, the larger the benefit.
- It might be more difficult to guarantee a smooth run on all types of machines (although the numba developers really try to make it work anywhere).


dlohmeier:
This is also a suggestion to handle issue #199.

dlohmeier:
I will create a new branch for further development.

PR #233, opened 2021-01-28
Add mass storage component

jkisse:
This is a simple storage component. Basically, it is like a sink (similar to the pandapower-storage) but I think it is good to have storage units grouped in a separate table.

For a reasonable simulation, one will have to add controllers that manage the limits and changes in the filling level. They have to be implemented separately, like in pandapower. An example is provided as a tutorial.

Please let me know if you'd prefer other (initial) columns in the storage_fluid-table.

Maybe components for thermal storage tanks will be added later, so I named this one `mass_storage`. It should be possible to use this in gas and water networks.

- [x] new component
- [x] new create_storage
- [x] test for create_storage
- [x] tutorial for a Storage Controller
- [ ] include in plotting (simple_plot)
- [x] include in documentation


pandapipes/component_models/storage_fluid_component.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class StorageFluid(ConstFlow):
+    """
+
+    """
+
+    @classmethod
+    def table_name(cls):
+        return "storage_fluid"
```
jkisse, line 16: change to mass storage

pandapipes/component_models/storage_fluid_component.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @classmethod
+    def sign(cls):
+        return 1
```
jkisse, line 20: i.e. signing convention of sinks. Charging = m_dot > 0, discharge = m_dot < 0

pandapipes/component_models/storage_fluid_component.py
```diff
@@ -0,0 +1,48 @@
...
+    def get_result_table(cls, net):
+        """Get results.
+
+        :param net: The pandapipes network
+        :type net: pandapipesNet
+        :return: (columns, all_float) - the column names and whether they are all float type. Only
+                if False, returns columns as tuples also specifying the dtypes
+        :rtype: (list, bool)
+        """
+        return ["mdot_kg_per_s"], True
```
jkisse, line 48: add m_stored_kg

jkisse:
This requires a controller which adjusts m_stored_kg according to the mass flow

Issue #188, opened 2021-01-04
Confusing error messages in case of non-convergence


dlohmeier:
I just realized that when a pipeflow does not converge, two diverging error messages appear:

pandapipes.pipeflow - WARNING: Maximum number of iterations reached but hydraulics solver did not converge.
UserWarning: Pipeflow converged, however, the results are phyisically incorrect as pressure is at the nodes [  237.   280.   301. ... 85568. 85569. 85593.] are negative

The first is just a logger warning, the second one is a UserWarning without raising an exception. It occurs in the junction component model upon result extraction. This issue raises the question how to deal with non-convergence. How should the results be written into the tables? Should the result tables be deleted, filled with NaN-values or just filled with whatever values calculated?

dlohmeier:
If the maximum number of iterations is reached, an exception should be raised. Results should be handled similar to pandapower. Maybe add a hint to the "_pit" entry in the logger warning.

dlohmeier:
In pandapower the following happens:

- init_results is done before the algorithm (i.e. empty result tables)
- if the algorithm did not converge, a LoadFlowNotConverged exception is raised, the result tables are not changed anymore
- result extraction happens after that check, i.e. in case of no convergence, no results are extracted

I think that is not a bad idea and we should implement this in pandapipes as well.

kbensafta:
Issue #337 seems to be on the same thread but ahead of this issue. Should therefore this issue be closed or transferred to discussion?

dlohmeier:
Thanks for the hint, the suggestions are already implemented, so I will close this issue.

PR #234, opened 2021-01-28
Adapt convergence

dlohmeier:
In case of non-convergence, now a PipeFlowNotConverged exception is raised within the pipeflow.


dlohmeier:
This should not be merged yet, as one more thing needs to be implemented to solve issue #188 : The result tables need to be initialized (empty or with nans) before the convergence check, so that they are all empty or contain nans if the pipeflow does not converge.

dlohmeier:
It seems like all changes were already merged in other PRs

PR #317, opened 2021-11-05
Adaption of std type functions

dlohmeier:
Related to Issue #312
- new functions copied from pandapower
- moved standard types from net.std_type to net.std_types
- moved StdType class to other module
- renamed some function args


dlohmeier:
Todos left:
- Change all std types such that they only contain relevant information for the network. The Pipe.csv table can be helpful and left the way it is, but when loading it to net.std_types, it should be converted rather than during the create-function.
- Pump std_types: Structure seems a little confusing and in the components, the calls are not vectorized.

jkisse (approved):
looks good to me

PR #318, opened 2021-11-06
The ordered values seem not to fit.

SimonRubenDrauz:
It seems that the from and to velocities do not fit the mean velocities. I think there is an error, also in one of the test seems an error.


SimonRubenDrauz:
wrong implementation

PR #371, opened 2022-06-14
enable using "run" parameter for run_control for multinet

rbolgaryn:
for run_control, one can pass "run" parameter to specify the function (e.g. pp.runpp, pp.rundcpp, pp.runopf, etc.).

For multinet, it becomes a dictionary with net names as keys and functions as values:

run={"power_net": pp.runpp, "power_net_dc": pp.rundcpp, "gas_net": ppipes.pipeflow}

Also, rearranged code to make it more compact by using .get for dictionaries


dlohmeier (approved):
Great work, thanks for the improvements and added tests!

PR #387, opened 2022-07-22
some bug fixes

dlohmeier:
- der_lambda could fail in some cases due to a bad correction term
- NetCalculationNotConverged error added to possible pandapipes errors in run_timeseries
- added some cases for numba calculations in _sum_by_group

### DIFF
--- a/pandapipes/pf/internals_toolbox.py
+++ b/pandapipes/pf/internals_toolbox.py
@@ -99,7 +99,7 @@ def _sum_by_group(use_numba, indices, *values):
     ind_dt = indices.dtype
     indices = indices.astype(np.int32)
     max_ind = max_nb(indices)
-    if max_ind < 1e5 and max_ind < 10 * len(indices):
+    if (max_ind < 1e5 or max_ind < 2 * len(indices)) and max_ind < 10 * len(indices):
         dtypes = [v.dtype for v in values]
         val_arr = np.array(list(values), dtype=np.float64).transpose()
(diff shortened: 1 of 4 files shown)


pandapipes/pf/internals_toolbox.py
```diff
@@ -99,7 +99,7 @@ def _sum_by_group(use_numba, indices, *values):
     ind_dt = indices.dtype
     indices = indices.astype(np.int32)
     max_ind = max_nb(indices)
-    if max_ind < 1e5 and max_ind < 10 * len(indices):
+    if (max_ind < 1e5 or max_ind < 2 * len(indices)) and max_ind < 10 * len(indices):
         dtypes = [v.dtype for v in values]
```
SimonRubenDrauz, line 103: Why is this required?

dlohmeier: it wouldn't make sense not to use numba in case of large grids. The condition should only check whether maximum index is much higher than the index length...

dlohmeier: probably, there is still a better way to do it, but since it is just a condition to check whether numba makes sense or not, it is not too important to have the optimal solution here.

SimonRubenDrauz: Okay, I see. Thanks for the clarification.

SimonRubenDrauz (approved):

PR #389, opened 2022-07-25
Stanet conversion

dlohmeier:
Still missing:

- FluidPropertySutherland get_at_integral_value
- valve_pipe component from STANET
- tests
- ...


dlohmeier:
In the tests, we get an import error for shapely, this needs to be fixed... so we want shapely not to be included necessarily, but only if the desired code is really run. We need to check if this is always necessary for the STANET converter.

PR #412, opened 2022-10-15
added valve_pipe component to stanet converter

dlohmeier:


pandapipes/converter/stanet/valve_pipe_component/create_valve_pipe.py
```diff
@@ -0,0 +1,94 @@
...
+from pandapower.auxiliary import get_free_id, _preserve_dtypes
+from pandapipes.converter.stanet.valve_pipe_component.valve_pipe_component import ValvePipe
+
+
+def create_valve_pipe(net, from_junction, to_junction, length_km, diameter_m, opened=True,
+                      loss_coefficient=0, sections=1, std_type=None, k_mm=0.15e-3,
+                      max_vdot_m3_per_s=100, max_v_m_per_s=None, alpha_w_per_m2k=0., qext_w=0.,
+                      name=None, index=None, in_service=True, type="valve_pipe", **kwargs):
+    """
+    Creates a valve pipe element in net["valve_pipe"] from valve pipe parameters.
```
jkisse, line 12: A little description how this differs from "pipe" and "valve" would be helpful (also for the autodoc). Maybe also, why this is hidden in the Stanet converter and should not be used otherwise

pandapipes/converter/stanet/valve_pipe_component/create_valve_pipe.py
```diff
@@ -0,0 +1,94 @@
...
+            calculations, where variables are dependent on pipe length.
+    :type sections: int, default 1
+    :param std_type:
+    :type std_type: str, default None
+    :param k_mm:
+    :type k_mm: float, default 0.15e-3
+    :param max_vdot_m3_per_s:
+    :type max_vdot_m3_per_s: float, default 100
+    :param max_v_m_per_s:
+    :type max_v_m_per_s: ?, default None
```
jkisse, line 39: What is the purpose of these parameters? If they are simply there to transfer information from the Stanet net to the respective table and otherwise ignored by the pipeflow etc. this should be stated here

jkisse:
I think `valve_pipe` should also be considered everywhere where all branch components are listed, e.g. in the toolbox [here](https://github.com/dlohmeier/pandapipes/blob/222e1d4d856429208aa6cc9614591f0295bb6586/pandapipes/toolbox.py#L123)
And does the if ... pipe clause [here ](https://github.com/dlohmeier/pandapipes/blob/222e1d4d856429208aa6cc9614591f0295bb6586/pandapipes/toolbox.py#L454) in `drop_elements_at_junction` need to be extended to `valve_pipe` as well?

pandapipes/converter/stanet/valve_pipe_component/create_valve_pipe.py
dlohmeier: I extended the docstrings.

pandapipes/converter/stanet/valve_pipe_component/create_valve_pipe.py
dlohmeier: @SimonRubenDrauz can you comment on that? I think that this is rather a remnant of the STANET model? Suppose it doesn't really add value to our model, does it?

Issue #361, opened 2022-03-10
FlowController for heat networks


TheNeedForSleep:
Hello,

Im currently trying to regulate flow at heat_exchangers. I have created an `MaxFlowController` that continuously reduces the diameter of an object until the flow is lower than specified. This sadly leads to errors  which i suspect is caused when very large values are next to small values.

Id like to have the possibility to control the max_flow in heat_exchangers similar to the way it works with sinks.
Sinks allow to specify a fixed maximum mass flow.

Example: I have a (small part of a) district heating network with one single producer and n consumer. One of the consumers is very close to the producer. The producer is located on the top right. The closest consumer is the first orange marker right next to it. ~75% of the flow goes through the first consumer while only ~25% goes to the the rest of the network which is not the desired behaviour.
![image](https://user-images.githubusercontent.com/44329937/157642288-dfbfc7d3-7677-4d33-b01f-0df9d5a3bb03.png)

Would it be possible to implement such a feature?
Either extend:
- valve with max_flow
- make sink have an out_flow
- extend heat_exchanger with a max_flow

Best regards

dlohmeier:
Hi @TheNeedForSleep ,
as @jkisse mentioned, there is a (still open) pull request for a new component, a flow control valve that exactly defines the flow going through. I think that would be a typical component for a heating grid. Is that what you were looking for? If you have any suggestions on how to adapt this model, please feel free to comment on the PR #413 . Otherwise we will merge it soon, but you could still make suggestions on improvement of course. Sorry for the late reply, but we have been busy on many gas grid related topics most of the time.
Best regards

PR #413, opened 2022-10-15
Flow control model

dlohmeier:
The flow controller is a branch component that controls the flow through itself. It means that the pressure drop is adapted to reflect the desired flow situation. This is especially helpful for heating grids, where otherwise the flow through heat exchangers needs to be controlled through the loss coefficient.
As the flow controller is a new component, it must be reflected in many parts of pandapipes:
- There is a new FlowControl component class (all other component models did not really change).
- There are new create-functions (single and bulk) in the create.py module.
- The flow controller was added to the plotting library (i.e. there is a new collection and a new patch function, and it was added to the simple_plot function call) as well as a "deafult" component to networkx graphs and toolbox functions.
- New tests were implemented to test the flow control behavior and other aspects like graph searches or plotting collections.


jkisse:
This might solve  #361

jkisse:
I think this will be very helpful in heating networks. But does this actually work with gas grids? If the velocity is fixed, the mass flow still depends on the density (but the user will assume that it is enforced, since it is an input parameter). I'd propose that an Error or Warning should be raised if `fluid.is_gas`

dlohmeier:
I think there is a misunderstanding here. We fix the velocity internally, that is true. But we have to keep in mind that this is the velocity at norm state, so in order to derive the mass flow from it, we will always multiply it with the norm density. The velocity itself will change indeed inside this component, depending on the pressure / compressibility.

dlohmeier:
I added a test for the flow control model and also added it to the toolbox functions and nx-graph creation. What is still missing is plotting. I would like to draw a valve patch with a T above to show that the valve is regulated.

dlohmeier:
Thanks for the great addition! Something that just crossed my mind is that we might need a keyword like "control_active" as in the pressure regulator component. If it is set to false, the flow control acts like an open valve. For plotting I thought of something fancy:
- if not "in_service": make it white
- if "in_service" but not "control_active": fill
- if "in_service" and "control_active": fill half (i.e. split the drawn triangles and make one filled and one white triangle out of it)

dlohmeier:
I added the control active behavior (actually the entry existed already, but was not considered). The plotting is solved as follows: the filling is set by in_service, the nice "controlling" fancy arrow is set by the control_active keyword. Or should this be changed to not confuse it with normal valves?

jkisse:
I think it would be nice to make it look slightly different to a valve, e.g. when someone wants to evaluate operational options based on a plot of the network.
What do you think about an "I" for inactive and "T" for active status (starting at the x of the valve symbol)?
Implementation could be in the FancyArrow by multiplying head_with and control_active:
`polys.append(FancyArrow(center[0], center[1], dx, dy, head_length=0, head_width=2*size*control_active)`

dlohmeier:
Thanks for the hint, I did that.

jkisse:
Heat exchanger and valve have a `loss_coefficient` - would it make sense to add a loss coefficent for the flow control as well?

dlohmeier:
I am not too sure about that because the flow controller models an ideal control valve. In case of active control (which should be the normal behavior), the loss coefficient would have no effect on the model. It only comes into play when the flow controller acts as an open valve.

pandapipes/component_models/abstract_models/branch_models.py
```diff
@@ -143,13 +143,13 @@ def adaption_before_derivatives_hydraulic(cls, net, branch_pit, node_pit, idx_lo
...
-    def calculate_temperature_lift(cls, net, branch_pit, node_pit):
+    def calculate_temperature_lift(cls, net, branch_component_pit, node_pit):
         """
 
         :param net:
         :type net:
-        :param branch_pit:
-        :type branch_pit:
+        :param branch_component_pit:
+        :type branch_component_pit:
```
EPrade, line 152: Looks good !

EPrade (dismissed):

pandapipes/component_models/flow_control_component.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        fc_pit = super().create_pit_branch_entries(net, branch_pit)
+        fc_pit[:, D] = net[cls.table_name()].diameter_m.values
+        fc_pit[:, AREA] = fc_pit[:, D] ** 2 * np.pi / 4
+        fc_pit[:, VINIT] = net[cls.table_name()].controlled_mdot_kg_per_s.values / \
+            (fc_pit[:, AREA] * fc_pit[:, RHO])
+
+    @classmethod
+    def adaption_before_derivatives_hydraulic(cls, net, branch_pit, node_pit, idx_lookups, options):
+        pass
```
EPrade, line 54: What is the use of this function?

dlohmeier: In general, this method is used by some components to model special behavior, e.g. a pressure lift for pumps. It is necessary, as this is not considered inside the global hydraulic derivative calculation otherwise. In case of the flow controller, this method is just skipped.

jkisse: Little addition: The method is called for all branch components during the pipeflow calculation, which is why it cannot be left away. With this "pass" implementation, it can be called, nothing happens, and that's fine ;-)

pandapipes/plotting/simple_plot.py
```diff
@@ -243,97 +279,71 @@ def create_simple_collections(net, respect_valves=False, respect_in_service=True
...
     if 'valve' in net:
         valve_colls = create_valve_collection(net, size=valve_size, linewidths=pipe_width,
                                               color=valve_color, respect_valves=respect_valves)
         collections["valve"] = valve_colls
 
+    if 'flow_control' in net:
+        flow_control_colls = create_flow_control_collection(
+            net, size=flow_control_size, linewidths=pipe_width, color=flow_control_color,
+            respect_in_service=respect_in_service)
+        collections["flow_control"] = flow_control_colls
```
SimonRubenDrauz, line 304: Why are you changing the structure here? Why is respect_in_service part of the function, whilst in all other cases it is not?

pandapipes/plotting/patch_makers.py
```diff
@@ -3,7 +3,7 @@
 # Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 import numpy as np
-from matplotlib.patches import RegularPolygon, Rectangle, Circle, PathPatch
+from matplotlib.patches import RegularPolygon, Rectangle, Circle, PathPatch, FancyArrow
```
SimonRubenDrauz, line 6: nice wording :)!

pandapipes/component_models/flow_control_component.py
```diff
@@ -0,0 +1,144 @@
...
+        :type net: pandapipesNet
+        :param branch_pit:
+        :type branch_pit:
+        :return: No Output.
+        """
+        fc_pit = super().create_pit_branch_entries(net, branch_pit)
+        fc_pit[:, D] = net[cls.table_name()].diameter_m.values
+        fc_pit[:, AREA] = fc_pit[:, D] ** 2 * np.pi / 4
+        fc_pit[:, VINIT] = net[cls.table_name()].controlled_mdot_kg_per_s.values / \
+            (fc_pit[:, AREA] * fc_pit[:, RHO])
```
SimonRubenDrauz, line 50: Why do we need no loss coefficient?

pandapipes/component_models/flow_control_component.py
```diff
@@ -0,0 +1,144 @@
...
+        # of velocity is allowed during the pipeflow iteration
+        f, t = idx_lookups[cls.table_name()]
+        fc_pit = branch_pit[f:t, :]
+        in_service_elm = np.isin(net[cls.table_name()].index.values,
+                                 fc_pit[:, ELEMENT_IDX].astype(np.int32))
+        active = net[cls.table_name()].control_active.values[in_service_elm]
+        fc_pit[active, JAC_DERIV_DP] = 0
+        fc_pit[active, JAC_DERIV_DP1] = 0
+        fc_pit[active, JAC_DERIV_DV] = 1
+        fc_pit[active, LOAD_VEC_BRANCHES] = 0
```
SimonRubenDrauz, line 68: This I do not get. I agree that the velocity is not changing. But LOAD_VEC_BRANCHES seems to me not a differential. Thus I would say, LOAD_VEC_BRANCHES is controlled_mdot_kg_per_s?

pandapipes/component_models/flow_control_component.py
```diff
@@ -0,0 +1,144 @@
...
+        # set all pressure derivatives to 0 and velocity to 1; load vector must be 0, as no change
+        # of velocity is allowed during the pipeflow iteration
+        f, t = idx_lookups[cls.table_name()]
+        fc_pit = branch_pit[f:t, :]
+        in_service_elm = np.isin(net[cls.table_name()].index.values,
+                                 fc_pit[:, ELEMENT_IDX].astype(np.int32))
+        active = net[cls.table_name()].control_active.values[in_service_elm]
+        fc_pit[active, JAC_DERIV_DP] = 0
+        fc_pit[active, JAC_DERIV_DP1] = 0
+        fc_pit[active, JAC_DERIV_DV] = 1
```
SimonRubenDrauz, line 67: m = rho * v * area --> 0 = rho * v * area - m --> df/dv = rho * area?

pandapipes/component_models/flow_control_component.py
dlohmeier: I always stumble over this part as well... the LOAD_VEC_BRANCH is the difference between the calculated pressure drop and the pressure drop given by the current estimates of the pressure at the connected nodes. Whereas the LOAD_VEC_NODES is the part that resumes over the mass flows coming into each node. So we assume that the difference is 0, because we accept any assumed pressure values at the connected nodes to be calculated correctly, we assume that the flow controller can produce any pressure drop / lift, that is required to create the desired mass flow.

pandapipes/component_models/flow_control_component.py
dlohmeier: Because the head loss is always a result of the flow which must be controlled. If we assume 0 length, we could output the loss coefficient as a result, but I am not sure if that is very interesting.

pandapipes/test/plotting/test_simple_collections.py
```diff
@@ -62,8 +63,15 @@ def test_simple_collections(base_net_is_with_pumps):
...
 
+
 def test_simple_collections_out_of_service(base_net_oos_with_pumps):
     net = copy.deepcopy(base_net_oos_with_pumps)
+
+    j11 = pandapipes.create_junction(net, 1.05, 293.15, name="Junction 11", geodata=(14, 0))
+    j12 = pandapipes.create_junction(net, 1.05, 293.15, name="Junction 12", geodata=(14, -2))
+    j10 = net.junction.loc[net.junction.name == "Junction 10"].index[0]
+    pandapipes.create_flow_controls(net, [j10, j10], [j11, j12], 0.5, 0.1, in_service=[True, False])
+
```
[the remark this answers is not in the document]
dlohmeier: Hmm... probably not a bad idea, but that would require another thought on how to adapt all other tests implementing this fixture and rename it in a suitable way. I opened an issue for that: #435.

pandapipes/create.py
```diff
@@ -743,22 +745,66 @@ def create_circ_pump_const_mass_flow(net, from_junction, to_junction, p_bar, mdo
...
+    (boost ratio)
+
+    :param net: The net for which this compressor should be created
+    :type net: pandapipesNet
+    :param from_junction: ID of the junction on one side which the compressor will be connected with
+    :type from_junction: int
+    :param to_junction: ID of the junction on the other side which the compressor will be \
+        connected with
+    :type to_junction: int
+    :param pressure_ratio:
```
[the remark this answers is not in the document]
dlohmeier: done

pandapipes/create.py
```diff
@@ -743,22 +745,66 @@ def create_circ_pump_const_mass_flow(net, from_junction, to_junction, p_bar, mdo
...
+    :type pressure_ratio: float
+    :param name: A name tag for this compressor
+    :type name: str, default None
+    :param index: Force a specified ID if it is available. If None, the index one higher than the\
+            highest already existing index is selected.
+    :type index: int, default None
+    :param in_service: True for in_service or False for out of service
```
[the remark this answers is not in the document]
dlohmeier: done

pandapipes/create.py
```diff
@@ -1175,7 +1277,7 @@ def create_pressure_controls(net, from_junctions, to_junctions, controlled_junct
     and 'to_junctions' must be arrays of equal length. Other parameters may be either arrays of the\
     same length or single values.
 
-    :param net: The net within this pump should be created
+    :param net: The net for which this pressure controls should be created
```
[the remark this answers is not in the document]
dlohmeier: done

pandapipes/create.py
```diff
@@ -818,6 +864,62 @@ def create_pressure_control(net, from_junction, to_junction, controlled_junction
     return index
 
 
+def create_flow_control(net, from_junction, to_junction, controlled_mdot_kg_per_s, diameter_m,
+                        control_active=True, name=None, index=None, in_service=True, type="fc",
```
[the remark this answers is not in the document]
dlohmeier: No, but that is the case in the pressure control. There are 3 possibilities as mentioned in the discussion:
- in_service and control_active: expected behavior
- in_service and not control_active: serves as an open valve (no flow control)
- not in_service (control_active not considered): serves as a closed valve
We can't put all aspects into one flag, and this one would be consistent with the pressure control.

pandapipes/plotting/simple_plot.py
jkisse: The implementation was more similar to the valve collections but I think it makes sense to align it with the other collections, for which in_service is considered prior to  the create...collection() calls. I've changed it.

pandapipes/create.py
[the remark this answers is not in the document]
SimonRubenDrauz: Agree, thanks. Forgot about that. Looks good to me then.

pandapipes/component_models/flow_control_component.py
SimonRubenDrauz: no, leave it like that. Thanks for the clarification.

pandapipes/component_models/flow_control_component.py
SimonRubenDrauz: Ahh, true, now I get it. Thanks!

pandapipes/component_models/flow_control_component.py
dlohmeier: Please note that we are looking at the branch_pit here. So, there we have the formulation for the pressure drop as a result of the velocity dependent friction. However, we are not interested into modeling this aspect here, so the only formulation here is that whatever velocity term we get from the node formulations, we accept (the load vector is 0). The formulation for the nodes (JAC_DERIV_DV_NODE) is not changed, the node balances are still part of the system matrix.

jkisse:
I've added a chapter in the documentation. Please check & update it if required, @dlohmeier

dlohmeier:
Looks good, thanks!

pandapipes/component_models/flow_control_component.py
SimonRubenDrauz: I see

SimonRubenDrauz (approved):

Issue #320, opened 2021-11-15
Version labelling of development branch (dev releases)


dlohmeier:
I came across a problem when adapting the `convert_format` function on the develop branch. Usually, we need the following code to not convert the format of a network with the same version:

    if isinstance(net.version, str) and version.parse(net.version) >= version.parse(__version__):
        return net

However, while working on this function in develop mode, while the pandapipes version has not changed, this will lead to failing tests. Thus, I would like to suggest to adapt the version of pandapipes on the develop branch. As the dev version is always the pre-release version of the next release, I would like to increment the release number and add a 'dev' identifier, but I think, this might have to be discussed first. This is how the packaging.version module works:

version.parse("0.5.0") < version.parse("0.5.1")  --> True
version.parse("0.5.0") < version.parse("0.5.1.dev_0")  --> True
version.parse("0.5.0") < version.parse("0.5.0.dev_0")  --> False
version.parse("0.5.1.dev_0") < version.parse("0.5.1") --> True

In my opinion, we should always work on the dev-version for the next release on our develop branch, and assume that it will be a minor release. In the release process, the version number has to be adapted when merging to the master branch.

dlohmeier:
For intermediate states, we can change the dev version:

version.parse("0.5.1.dev_0") < version.parse("0.5.1.dev_1") --> True

dlohmeier:
I think, this is done now. Can we close this issue?

dlohmeier:
This should now be solved through PR #418. The following rules apply (as far as I understand, please correct me if I'm wrong @SimonRubenDrauz) :
- The first commit on develop will increase the net version to x.x.x.postx (e.g. from 0.8.1 for the release to 0.8.1.post1).
- The format version increases only when the format changes. On the one hand, this means that in net version 0.11.2 there could still be format version 0.8.1. On the other hand, the net version could be 0.8.1.post1 and the format version 0.8.1.post3.
- A new release should increase the net version to the new release tag (i.e. from 0.8.1.post1 to 0.8.2 or 0.9.0), while the format version is only addressed if in post-mode (i.e. from 0.8.1.post4 to 0.9.0, but if 0.8.0, it will remain the same).

kbensafta:
Should this issue be considered as closed?

PR #418, opened 2022-10-20
Format version

SimonRubenDrauz:


dlohmeier (approved):
Perfect, that's exactly what we need!

dlohmeier:
This should solve #320 and is related to pandapower issue e2nIEE/pandapower#1421.

PR #420, opened 2022-10-21
toolbox function to extract the pit structure as pandas tables

dlohmeier:


jkisse (approved):
nice feature!

dlohmeier:
I am currently writing a test for this function. Unfortunately, I found that there is a special network in our fixtures that does not converge, although it should. Thus, it will take another little while, but in the end, there should be way better test coverage.

dlohmeier:
Unfortunately, I couldn't solve the problem, but I chose a different testing network for this use case.

PR #421, opened 2022-10-21
Format version

SimonRubenDrauz:
bug detected

### DIFF
--- a/pandapipes/io/convert_format.py
+++ b/pandapipes/io/convert_format.py
@@ -23,7 +23,7 @@ def convert_format(net):
     format_version = version.parse(__format_version__)
     # For possible problems with this line of code, please check out
     # https://github.com/e2nIEE/pandapipes/issues/320
-    if not hasattr(net, 'format_version'):
+    if version.parse(net.version) <= version.parse('0.7.0'):
         net.format_version = net.version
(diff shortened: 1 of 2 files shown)


dlohmeier (commented):
What errors would occur there? I didn't quite get why the previous version was wrong...

SimonRubenDrauz:
> What errors would occur there? I didn't quite get why the previous version was wrong...

The reason is the following. As soon as you load a network, an empty network is created and the different parameter are passed afterwards. If you have a net, which doesn't have the attribute format_version, it is not overwritten. Thus, the current format_version remains and therefore convert_format won't be applied. The reason, why the tests passed before, was a typo: __format_version__ instead of format_version. 

Do you have any other idea what might be a better way to solve that problem.

dlohmeier:
That's indeed an issue, and I think that we should rather try to make adaptions to the `from_json()` part. Does it appear in other formats as well? The correct behavior, whether with or without `convert_format()`, should be that we load the old net and if format_version doesn't exist, we replace this value with the saved net version. But this has to be done before the `convert_format()` task, of course.

SimonRubenDrauz:
adapted it accordingly. Please check it out if you agree with the solution.

dlohmeier (approved):
Looks good!

PR #427, opened 2022-11-18
variable encodings in read_option

nsanina:
set the encoding typ variable to adjust in the read_options. If nothing is given encoding = "cp1252"

@dlohmeier please review this changes


dlohmeier (approved):
Thanks for the change!

PR #429, opened 2022-11-21
change k_mm in the create_pipes functions to 0.2 mm instead of 1 mm

### LINKED ISSUE #323: default k value is too big
k is by default equals to 1mm. That is quite unrealistic. 0.2mm would be much better.

jkisse:
0.2 mm is a more realistic than 1 mm. It is still a quite high value
closes #323 
change level: minor - should not be merged if the next release is just a bugfix release

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 Change Log
 =============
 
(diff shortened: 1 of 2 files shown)


SimonRubenDrauz (approved):
Looks good!

PR #430, opened 2022-11-22
add simple_plotly function for pandapipes

jkisse:
Like `simple_plotly` from pandapower but for pandapipes (interactive plotting function).
So far, only tested with gas grids.
Some components are still missing (heat exchanger, compressor, valve, different pump types) and the infofunction could be improved.
Contributions welcome!


jkisse:
rename to simple_plotly_gas

jkisse:
Traces for compressors and valves added (as lines with dedicated colors)
simple_plotly extended respectively. Closed valves are represented as dotted lines.

jkisse:
I won't be able to add heat grid elements to plotly, but I'd like to merge this feature (at least for gas) soon. Can we add a disclaimer that it only makes sense for gas networks so far? (e.g. a check if the fluid is a gas)

dlohmeier:
Seems fine to me.

PR #483, opened 2023-01-05
Hotfix

SimonRubenDrauz:
- shapely as required package added as it is required for the stanet converter


jkisse (dismissed):

dlohmeier (approved):
I think it's indeed better to make shapely a requirement, otherwise the STANET converter cannot be used with a normal installation.

PR #497, opened 2023-01-27
Flow control is considered in nx_graph creation

dlohmeier:
Hotfix for nx_graph creation of networks with flow controllers, as the keywords concerning them were not correctly considered within the function (include_flow_controls could not be set to False).


dlohmeier:
I chose the wrong base branch.

PR #505, opened 2023-01-31
Hotfix/flow ctrl nxgraph

dlohmeier:
I added a hotfix for the nxgraph creation, as flow controllers are not considered correctly there.


SimonRubenDrauz:
Could you move the hotfix-branch directly to the e2niee/pandapipes repo, so that I can also add my hotfix?

PR #507, opened 2023-02-01
bugfix if all pumps are out of service

SimonRubenDrauz:


dlohmeier (approved):
Hard to believe, we missed this until now :D

dlohmeier:
Should this be added to our hotfix list? Maybe, we can still add this to #505 ?

PR #524, opened 2023-03-13
remove Python 3.7 from test workflow

### LINKED ISSUE #499: Drop support for Python 3.7
Pandapower is going to drop Python 3.7 soon, cf. [Issue 1841](https://github.com/e2nIEE/pandapower/issues/1841) 
Pandas and numpy already stopped support for Python 3.7

For easier maintenance, pandapipes should also stop support for Python 3.7 (i.e., remove it from the test workflow)

jkisse:
Python 3.7 tests are currently failing and there is not really a need to fix them. This PR adapts the GitHub Actions script so that Python 3.7 will no longer be tested.
closes #499

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,7 +75,7 @@ jobs:
     needs: upload
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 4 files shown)


dlohmeier (approved):
Since pandapower now introduces python 3.8 specific implementations (like string formatting notation), this makes sense and should encourage to adapt such kind of notations, as they are easier to implement and read.

PR #525, opened 2023-03-17
Biomethane

kbensafta:
This PR is to add biomethane to pandapipes. A particularity of injecting biomethane in the gas grid in Germany is the addition of LPG to enhance the Wobbe-Index and the heating value of the gas and thus make it equivalent to the natural gas transported in the grid. This addition is done in a gas treatment unit (Biogaseinspeiseanlage) upstream of the biomethane feed-in junction.

As a consequence this PR contains two biomethane compositions:

- The folder _biomethane_  represents the composition and the properties of the fluid accroding to the norm (Arbeitsblatt DVGW G 260(A)) upstream of gas treatment unit.
- The folder _biomethane_LPG_  represents the composition and the properties of the fluid accroding to the norm (Arbeitsblatt DVGW G 260(A)) downstream of the treatment unit, i.e., at the feed-in junction.

Both compositions could be considered as relevant. _Biomethane_LPG_ represents the composition of the "currently" fed-in gas.
As to _Biomethane_, it represents how the gas will be fed-in in the future, since the gas DSOs expect to not enrich the biomethane with LPG anymore. To insure a proper gas billing, they intend to use modeling and simulation to determine the heating value of the gas mixture (biomethane and natural gas) transported through the network.

**Should both gas compositions be added to pandapipes ?**

The properties of the biomethane(s) are computed based on the package [CoolProp](http://www.coolprop.org/).


dlohmeier:
To me this sounds good and very valuable. Some things that would be very helpful to add:
- add the fluids to the fluid library (call_lib in properties/fluids.py)
- a description how the values are derived and how they are defined (would probably make sense in the fluid library as well, or just somewhere inside the documentation)
- I would change the names to biomethane_pure and biomethane_treated, but that's just a suggestion.

kbensafta:
I have messed up the git history of this branch. I am closing this PR and opening a new one.

PR #530, opened 2023-04-24
Add dhnx pandapipes tutorial

joroeder:
This PR adds a tutorial showing a workflow for combining the DHNx package [https://github.com/oemof/DHNx](https://github.com/oemof/DHNx) and pandapipes.

In the first part of the tutorial the network layout and dimensioning for a new district heating network is set up with DHNx (based on a Mixed-Integer linear optimisation), and then, a thermo-hydraulic simulation is performed with pandapipes for checking the technical feasibility of the district heating network. This tutorial might be very helpful for pandapipes users if no data about district heating networks is available or if new district heating networks are investigated.

I planned to convert the script into a jupyter notebook as all the other tutorials after a first feedback. What do you think about the tutorial?

Some remarks:
 * The tutorial got very long as I wanted to explain the process especially of the DHNx part in detail.
 * The folders and files in `invest_data` will become default values in future. So the only input data will be the `Pipe_data.csv`.


dlohmeier:
Hi @joroeder ,
thanks a lot for this great work! I think that this is extremely valuable and possibly just a first step for an even better interface between the two tools to create one easy-to-use open-source toolbox for district heating network dimensioning and combine the strengths of the two tools.
I had a first glimpse at your tutorial and have a few small questions:
- How big is the network you create? I would usually suggest to always use the "plural" or "bulk" create functions of pandapipes in cases where network data is already given as a table that just needs to be converted to be readable by pandapipes (i.e. create_junctions instead of create_junction with a for-loop etc.).
- How are district heating network sinks and sources modeled in DHNx? Usually, in pandapipes we try to encourage to create a closed-loop network that is fed by a circulation pump which connects flow and return system, which are on the customer side connected via flow controls and heat exchangers to ensure the correct heat extraction. On the other hand, sinks and sources are units that extract or feed in mass flows to the system, which usually makes the thermal model hard to control. Does the DHNx model work completely different?

I think we would still have to clarify how to integrate this tutorial within pandapipes. Some thoughts:
- Regarding the input data that you create with the help of osmnx, I am wondering how to handle the workflow of your tutorial such that it can also be used offline. Maybe some additional default files would be helpful.
- You use packages that are not part of pandapipes, which is completely fine for an advanced tutorial, but I am not sure how we handled such cases before. We have a tutorial check that ensures that all tutorials are running in new releases, so in this case either we need to extend the github actions' package list, or exclude the tutorial from the check (also maybe because of downloading data from osmnx through github actions could be problematic).

joroeder:
Hi @dlohmeier, thank you for your feedback!

> 
> * How big is the network you create? I would usually suggest to always use the "plural" or "bulk" create functions of pandapipes in cases where network data is already given as a table that just needs to be converted to be readable by pandapipes (i.e. create_junctions instead of create_junction with a for-loop etc.).

The example is not very big, and if there are proper function to create the network, we of course should replace the for loops. Actually, until this example, I did not work with pandapipes, but a Masters student of mine.

> 
> * How are district heating network sinks and sources modeled in DHNx? Usually, in pandapipes we try to encourage to create a closed-loop network that is fed by a circulation pump which connects flow and return system, which are on the customer side connected via flow controls and heat exchangers to ensure the correct heat extraction. On the other hand, sinks and sources are units that extract or feed in mass flows to the system, which usually makes the thermal model hard to control. Does the DHNx model work completely different?

I don't think so. However, the modeling detail of the thermo-hydraulic is simplified, actually not existent in the DHNx optimisation. In DHNx, a linear mixed-integer investment optimization is performed, that only knows about loads and costs for transport capacity. The DHS pipelines are considered to be forward and return in one with a thermal transport capacity (which requires some assumptions such as temperature spread. A mass flow based approach would also be possible.). So, actually, the model is very simple, the Sinks are nodes with outflow (which itself can be constrained), and the sources are nodes with one inflow. And the optimisation problem is concerned with finding the cheapest supply route. So, for the parametrization of the model the thermal and hydraulic physic needs to be pre-calculation. The actual dimensioning is based on a maximum pressure drop per meter, that determines the transport capacity of a pipelines trench. In the end, DHNx gives a first estimation about the routing and dimensions, that can then be further optimized with a detailed simulation with pandapipes. It might be also possible to apply this approach to gas networks ...

Regarding the loops, it guess it is no problem to add the return pipelines and substations (that require some further assumption such as pressure drop at the substations and return temperature on the secondary side of the consumers heating system). My idea was that a calculation of the forward piping systems gives a first insight into the feasibility of the drafted DHS network to evaluate temperature loss and pressure drop. The pressure drop of the returning piping system is in many cases almost the same, right? So, together with the maximum pressure drop at the substations of the TAB, the overall pressure drop can be estimated. For a closed-loop network, you also need to add the heat exchangers of the boilers, heat pumps, with their pumps and so on (also alle Erzeugerkreise), with all the hydraulic and controls of the heat supply site, right? Is this all possible with pandapipes? This would be very nice to provide such an extensive example of a heat supply site, but I am wondering if this is an overshoot for this example.

> 
> 
> I think we would still have to clarify how to integrate this tutorial within pandapipes. Some thoughts:
> 
> * Regarding the input data that you create with the help of osmnx, I am wondering how to handle the workflow of your tutorial such that it can also be used offline. Maybe some additional default files would be helpful.

Osmnx has the option of caching the data. However, then, the cached files are part of the repository. What do you mean by additional default files? Actually, my philosophy was to reduce the files to a minimum to keep the repository lean.

> 
> * You use packages that are not part of pandapipes, which is completely fine for an advanced tutorial, but I am not sure how we handled such cases before. We have a tutorial check that ensures that all tutorials are running in new releases, so in this case either we need to extend the github actions' package list, or exclude the tutorial from the check (also maybe because of downloading data from osmnx through github actions could be problematic).

Yes, this is a problem. We could also add an try ... except statement or something for the imports of the tutorial. so that the checks pass even if the additional packages are not installed. Okay, maybe this is the same as excluding the tutorial from the checks :D ... or maybe add `extra_requires` the profile `tutorial`, that contains all additional packages of the repository. A kind of extra examples/tutorials repository for examples connecting different packages could solve this problem. However, it is also nice if the tutorials are directly located within the repo, otherwise these are maybe more difficult to find.  ... just some thoughts

PR #531, opened 2023-04-26
Performance and convergence improvements

dlohmeier:
This PR is meant to merge changes from develop and to increase the performance through 2 simple steps:

- removal of a large logging call in the heat transfer simulation of pipeflow
- correction of the calculation of JdT and JdT1 in branch_model in transient mode

PR #532, opened 2023-04-26
 Performance and convergence improvements #531 

dlohmeier:
This PR is meant to merge changes from develop and to increase the performance through 2 simple steps:

- removal of a large logging call in the heat transfer simulation of 
- correction of the calculation of JdT and JdT1 in branch_model in transient mode


pandapipes/component_models/abstract_models/branch_models.py
```diff
@@ -130,26 +130,23 @@ def calculate_derivatives_thermal(cls, net, branch_pit, node_pit, idx_lookups, o
 
         transient = get_net_option(net, "transient")
 
-        tvor = branch_pit[:, T_OUT_OLD]
+        tvor = branch_component_pit[:, T_OUT_OLD]
```
EPrade: Can u explain this change? I do not quite understand why we can just change branch_pit to branch_component_pit

pandapipes/component_models/abstract_models/branch_models.py
```diff
@@ -130,26 +130,23 @@ def calculate_derivatives_thermal(cls, net, branch_pit, node_pit, idx_lookups, o
...
-
-            branch_component_pit[:, JAC_DERIV_DT_NODE] = rho * v_init \
-                                                         * branch_component_pit[:, AREA]
-            branch_component_pit[:, LOAD_VEC_NODES_T] = rho * v_init \
-                                                        * branch_component_pit[:, AREA] * t_init_i1
+                -(rho * area * cp * (t_m - tvor) * (1 / delta_t) * length + rho * area * cp * v_init * (
+                            -t_init_i + t_init_i1 - tl)
+                  - alpha * (t_amb - t_m) * length + qext)
+
+            branch_component_pit[:, JAC_DERIV_DT] = - rho * area * cp * v_init
```
EPrade: Why can we just drop the alpha part here?

pandapipes/component_models/abstract_models/branch_models.py
```diff
@@ -130,26 +130,23 @@ def calculate_derivatives_thermal(cls, net, branch_pit, node_pit, idx_lookups, o
...
-
-            branch_component_pit[:, JAC_DERIV_DT] = - rho * area * cp * v_init / length + alpha \
-                                                    + rho * area * cp / delta_t
-            branch_component_pit[:, JAC_DERIV_DT1] = rho * area * cp * v_init / length + 0 * alpha \
-                                                     + rho * area * cp / delta_t
-
-            branch_component_pit[:, JAC_DERIV_DT_NODE] = rho * v_init \
-                                                         * branch_component_pit[:, AREA]
-            branch_component_pit[:, LOAD_VEC_NODES_T] = rho * v_init \
-                                                        * branch_component_pit[:, AREA] * t_init_i1
```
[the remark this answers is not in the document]
dlohmeier: The node derivatives and load vector are calculated later again, as they are equal for transient and stationary simulations.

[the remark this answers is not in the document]
dlohmeier: so, I don't want to calculate them twice...

pandapipes/component_models/abstract_models/branch_models.py
dlohmeier: branch_component_pit is related to the specific component (e.g. pipe vs. valve), while branch_pit contains entries for all components. We only want to do the calculation for the single component here. A global formulation would be nice to have, as it is more performant, but has not been implemented yet.

pandapipes/component_models/abstract_models/branch_models.py
dlohmeier: That's the main reason for this PR, the correction of the derivatives. Since the heat losses are calculated based on the outlet temperature (I did this explicitly in another commit, but here you can see that tm=t_init_i1), it cannot be part of the partial derivative for the inlet temperature, it just gets 0 there.

dlohmeier:
@EPrade do you think the PR is valid? Or would you like to see some changes?

EPrade (approved):

PR #537, opened 2023-05-22
add diagnostic function

jkisse:
Add diagnostic functions to run some common checks. The function is intended to be applied if a pipeflow calculation does not converge.

Please feel free to modify and extend the function!

### DIFF
--- a/pandapipes/diagnostic.py
+++ b/pandapipes/diagnostic.py
@@ -1,0 +1,145 @@
+# Copyright (c) 2020-2023 by Fraunhofer Institute for Energy Economics
+# and Energy System Technology (IEE), Kassel, and University of Kassel. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+
+import pandapipes as pp
+import numpy as np
+
+from pandapipes import PipeflowNotConverged
+
+try:
+    import pandaplan.core.pplog as logging
+except ImportError:
+    import logging
+
+logger = logging.getLogger(__name__)
+
+
+def check_net(net, low_length_limit_km=0.01, check_scaling_factor=1e-5):
+    """
+    Run some diagnostic checks on the net to identify potential flaws.
+    """
+    net = net.deepcopy()  # do not modify the direct input
+    try:
+        pp.pipeflow(net)
+        if net.converged:
+            logger.info("The initial, unmodified pipeflow converges.")
+        else:
+            logger.warning("The initial, unmodified pipeflow does NOT converge.")
+    except Exception as e:
+        logger.info(f"The initial, unmodified pipeflow does NOT converge.\n"
+                    f"\t\tThis exception is raised:\n\t\t{e}")
+
+    # check ext_grid
+    if net.fluid.is_gas & (not hasattr(net, "ext_grid") | net.ext_grid.empty):
+        logger.warning("The net does not have an external grid! "
+                       "An external grid is required for gas networks.")
+
+    # check zero / low length
+    zl = net.pipe.loc[net.pipe.length_km == 0]
+    ll = net.pipe.loc[net.pipe.length_km <= low_length_limit_km]
+    if not zl.empty:
+        logger.warning(f"{len(zl.index)} pipes have a length of 0.0 km. (IDs: {zl.index})")
+    if not ll.empty:
+        logger.warning(f"{len(ll.index)} pipes have a length below"
+                       f" {low_length_limit_km} km. "
+                       f"This could lead to convergence issues. The lowest length in the net is "
+                       f"{ll.length_km.min()} km.\n"
+                       f"(IDs of pipelines with low length: {ll.index})")
+
+    net2 = net.deepcopy()
+    net2.pipe.loc[net2.pipe.length_km < low_length_limit_km].length_km = low_length_limit_km
+    try:
+        pp.pipeflow(net2)
+        if net2.converged:
+            logger.info(f"If all short pipelines (< {low_length_limit_km} km) were set to "
+                        f"{low_length_limit_km} km, the pipeflow would converge.")
+        else:
+            logger.warning(f"If all short pipelines (< {low_length_limit_km} km) were set to "
+                        f"{low_length_limit_km} km, the pipeflow would still NOT converge.")
+    except Exception as e:
+        logger.info(f"Pipeflow does not converge, even if all short pipelines (< 10 m) were set "
+                    f"to {low_length_limit_km} km. \n"
+                    f"\t\tThe error message is: {e}")
+
+    # check iterations
+    iter = 200
+    try:
+        pp.pipeflow(net, iter=iter)
+        logger.info(f"The pipeflow converges after {net._internal_results['iterations']:d} "
+                    f"iterations.")
+    except PipeflowNotConverged:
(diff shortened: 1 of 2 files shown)


src/pandapipes/diagnostic.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    Run some diagnostic checks on the net to identify potential flaws.
+    """
+    net = net.deepcopy()  # do not modify the direct input
+    try:
+        pp.pipeflow(net)
+        if net.converged:
+            logger.info("The initial, unmodified pipeflow converges.")
+        else:
+            logger.warning("The initial, unmodified pipeflow does NOT converge.")
```
dlohmeier, line 28: To me, it seems like this is unreachable code, as the pipeflow would raise an exception here which is only caught in the next line.

src/pandapipes/diagnostic.py
```diff
@@ -0,0 +1,145 @@
...
+
+try:
+    import pandaplan.core.pplog as logging
+except ImportError:
+    import logging
+
+logger = logging.getLogger(__name__)
+
+
+def check_net(net, low_length_limit_km=0.01, check_scaling_factor=1e-5):
```
dlohmeier, line 18: To me, the value of 0.01 km seems a bit high. I believe that this code was written when the unit was still m. So, we might rather change it to 1e-5.

src/pandapipes/diagnostic.py
```diff
@@ -0,0 +1,145 @@
...
+    try:
+        pp.pipeflow(net2)
+        if net2.converged:
+            logger.info(f"If all short pipelines (< {low_length_limit_km} km) were set to "
+                        f"{low_length_limit_km} km, the pipeflow would converge.")
+        else:
+            logger.warning(f"If all short pipelines (< {low_length_limit_km} km) were set to "
+                        f"{low_length_limit_km} km, the pipeflow would still NOT converge.")
+    except Exception as e:
+        logger.info(f"Pipeflow does not converge, even if all short pipelines (< 10 m) were set "
```
dlohmeier, line 61: make "10m" a variable

src/pandapipes/diagnostic.py
```diff
@@ -0,0 +1,145 @@
...
+        net3.sink.scaling *= check_scaling_factor
+    if hasattr(net, "source"):
+        net3.source.scaling *= check_scaling_factor
+    try:
+        pp.pipeflow(net3)
+        if net3.converged:
+            logger.info(f"If sinks and sources were scaled with a factor of to "
+                        f"{check_scaling_factor}, the pipeflow would converge.")
+        else:
+            logger.warning(f"If sinks and sources were scaled with a factor of to "
```
dlohmeier, line 87: See comment above.

src/pandapipes/diagnostic.py
```diff
@@ -0,0 +1,145 @@
...
+        logger.info(f"Pipeflow does not converge with sinks/sources scaled by"
+                    f" {check_scaling_factor}.\n"
+                    f"\t\tThe error message is: {e}")
+
+    # check k
+    if any(net.pipe.k_mm > 0.5):
+        logger.warning(f"Some pipes have a friction factor k_mm > 0.5 (extremely rough). The "
+                       f"highest value in the net is {net.pipe.k_mm.max()}. Up to "
+                       f"0.2 mm is a common value for old steel pipes."
+                       f"\nRough pipes: {net.pipe.loc[net.pipe.k_mm > 0.5]}.")
```
dlohmeier, line 99: I would not print out the whole dataframe, but rather just the index.

src/pandapipes/diagnostic.py
```diff
@@ -0,0 +1,145 @@
...
+    net4 = net.deepcopy()
```
dlohmeier, line 100: Is net.deepcopy() the same as copy.deepcopy(net)?

src/pandapipes/diagnostic.py
```diff
@@ -0,0 +1,145 @@
...
+    net4.pipe.k_mm = 1e-5
+    try:
+        pp.pipeflow(net4)
+        if net4.converged:
+            logger.info(f"If the friction factor would be reduced to 1e-5 for all pipes, "
+                        f"the pipeflow would converge.")
+        else:
+            logger.warning(f"If the friction factor would be reduced to 1e-5 for all pipes, "
```
dlohmeier, line 108: See comment above. --> "was reduced"

src/pandapipes/diagnostic.py
```diff
@@ -0,0 +1,145 @@
...
+            if len(missing):
+                logger.warning(f"Some {nc}s are connected to non-existing junctions!"
+                               f"\n{nc}s:{net[nc].loc[net[nc].junction.isin(missing)]}"
+                               f"\nmissing junctions:{missing}")
+
+    # check from and to junctions
+    branch_component = ["pipe", "valve", "compressor", "pump", "heat_exchanger", "circulation_pump"]
+    for bc in branch_component:
+        if hasattr(net, bc):
+            missing_f = np.setdiff1d(net[bc].from_junction, net.junction.index)
```
dlohmeier, line 128: I would prefer the following:
1. Add some code above that checks all tables in net and makes sure that the respective model is part of the component list (an important source of errors).
2. Iterate over all branch components and use the "from_to_junction" identifiers to check the columns

src/pandapipes/diagnostic.py
```diff
@@ -0,0 +1,145 @@
...
+            missing_t = np.setdiff1d(net[bc].to_junction, net.junction.index)
+            if len(missing_t) | len(missing_t):
+                logger.warning(f"Some {bc}s are connected to non-existing junctions!")
+                logger.warning(f"missing 'from' junctions:{missing_f}")
+                logger.warning(f"missing 'to' junctions:{missing_t}")
+
+
+if __name__ == '__main__':
+    import pandapipes.networks
```
dlohmeier, line 137: I think, we should rather skip main calls

src/pandapipes/toolbox.py
```diff
@@ -615,3 +615,40 @@ def get_internal_tables_pandas(net, convert_types=True):
                         tbl[col] = tbl[col].astype(np.bool_)
 
     return node_table, branch_table
+
+
+def print_pf_summary(net):
```
dlohmeier, line 620: We would need some kwargs here (e.g. for heating grids)

dlohmeier: We could also just choose between different modes here. But for a start, we can also just add a hint that this function returns some unuseful stuff for heating grids.

src/pandapipes/diagnostic.py
EPrade: You mean iterations as a function argument?

dlohmeier: I mean the pipeflow exception that must be caught. The else-part will never be reached.

jkisse:
I just had an interesting network that converged with alpha = 0.3 (nonlinear = "automatic") but not with alpha = 0.2 or alpha = 0.4.
Therefore, I've added an alpha_sweep function that tries 10 different alphas in the pipeflow.

src/pandapipes/diagnostic.py
jkisse: done

PR #538, opened 2023-05-31
adapted the heat exchanger creation in the stanet converter

dlohmeier:
- if required by user, the flow through heat exchangers is stored in the pandapipes net
- added qext_w from stanet
- no filtering for connections in heat exchangers (not clear what that was supposed to do, but created errors)

- additionally save temperature results for stanet nodes


pandapipes/converter/stanet/table_creation.py
```diff
@@ -574,27 +578,26 @@ def create_geodata_sections(row):
     )
 
 
-def create_heat_exchangers(net, stored_data, connection_table, index_mapping, add_layers):
+def create_heat_exchangers(net, stored_data, index_mapping, add_layers, add_flow=False):
```
jkisse, line 581: Wouldn't it be better to choose a different name to avoid confusion with create_heate_xchangers from create.py?
e.g. create_heat_exchangers_stanet

dlohmeier: Not a bad idea, I did rename it.

PR #543, opened 2023-06-19
fix: consider ambient pressure in calc. of compression work

### LINKED ISSUE #542: Isentropic compression work for a pump is not calculated over absolute pressures
**Describe the bug**
I´m using the pump component P2 with a gas (methane). A parallel calculation of the isentropic compressión power showed me that the results from the pump considered a pressure ratio computed wrongly over the manometric pressures.

**To Reproduce**
A very simple case code is attached to reproduce the bug. 
[test-isentropic-compression-methane.py.zip](https://github.com/e2nIEE/pandapipes/files/11780642/test-isentropic-compression-methane.py.zip)

**Expected behavior**
The expected value of the isentropic power of compression is 0.00683022 MW but the pandapipes results table reports 0.007752 MW.

**Python environment (please complete the following information):**
 - OS: MacOS Big Sur (11.7.6)
 - pandapipes version: 0-8-4

**Additional context**
I'm interested in using the pump component because I have the characteristic curve of the compressor.

jkisse:
fixes #542 

Please do not merge until we discussed the release plan


pandapipes/component_models/pump_component.py
```diff
@@ -150,8 +150,8 @@ def extract_results(cls, net, options, branch_results, nodes_connected, branches
             f, t = get_lookup(net, "branch", "from_to")[cls.table_name()]
             res_table = net["res_" + cls.table_name()]
             if net.fluid.is_gas:
-                p_from = branch_results["p_from"][f:t]
```
dlohmeier, line 153: Did we also produce an error with respect to compressibility at the nodes, since we did not consider the absolute pressure? (cf. line 159)

jkisse: Thanks for pointing this out. The compressibility is calculated on relative pressure, so the ambient pressure needs to be subtracted. I'll fix it

jkisse:
tests are failing because pandapower removed support for Python 3.7. If we add pandapower==2.11.1 in the setup.py, the tests will work with Python 3.7, but users with Python > 3.7 will not be able to use higher pandapower versions.

I think the removal of Python 3.7 should be at least a minor release, not a hotfix.

dlohmeier:
We will need to fix this in any case, and I think that it's OK to restrict pandapower to lower versions. We should keep in mind that all our previously released versions cannot be used with pandapower >2.11 anymore, at least not if we want to ensure that the tests can be run. If someone uses pandapipes 0.8.4 with pandapower 2.13.1, he will silently accept that the tests fail. Is that a desired behavior? I guess that limiting to pandapower < 2.12 is a reasonable option, as we want the two packages to go along.

pandapipes/component_models/pump_component.py
kbensafta: For my own understanding of the matter; why is the compressibility calculated for relative pressure ?

kbensafta: FYI, on Line 80:
`
p_from = node_pit[from_nodes, PAMB] + node_pit[from_nodes, PINIT]`

jkisse: It should be calculated at abs. pressure, thanks for pointing that out

dlohmeier: Again for my understanding: There was an error in the calculation before, right?

jkisse: The error in the int ital calculation was that `p_from` and `p_to` were relative pressures but handled like they were absolute pressures.  Sorry for the confusion...

dlohmeier (approved):
Looks good to me!

dlohmeier:
Could we please add PR #548 here?

PR #549, opened 2023-07-20
Link license

SteffenMeinecke:
advantage by this PR: no two different license versions possible


dlohmeier:
This looks like a great idea, it just seems that the path is not yet correct, or the docs build does something unexpected. Can you give some insights on that, @jkisse ?

SteffenMeinecke:
that is correct, the link was wrong

jkisse (approved):
Looks good, thank you!

PR #561, opened 2023-09-06
internal component arrays

dlohmeier:
- added "components" dict of arrays to pit
- every component can create its own internal array if needed (create_component_array in base_component.py)
- moved PRESSURE_RATIO and STD_TYPE from branch_idx.py to internal arrays of pumps and compressors, as they are only needed right there

### DIFF
--- a/pandapipes/component_models/component_toolbox.py
+++ b/pandapipes/component_models/component_toolbox.py
@@ -198,3 +198,23 @@ def standard_branch_wo_internals_result_lookup(net):
         required_results_hyd.extend([("v_mean_m_per_s", "v_mps")])
 
     return required_results_hyd, required_results_ht
+
+
+def get_component_array(net, component_name, internal=True):
+    """
+    Returns the internal array of a component.
+
+    :param net: The pandapipes network
+    :type net: pandapipesNet
+    :param component_name: Table name of the component for which to extract internal array
+    :type component_name: str
+    :param internal: If True, only return entries of active elements (included in _active_pit)
+    :type internal: bool
+    :return: component_array - internal array of the component
+    :rtype: numpy.ndarray
+    """
+    f_all, t_all = get_lookup(net, "branch", "from_to")[component_name]
+    if not internal:
+        return net["_pit"]["components"][component_name]
+    in_service_elm = get_lookup(net, "branch", "active_hydraulics")[f_all:t_all]
+    return net["_pit"]["components"][component_name][in_service_elm]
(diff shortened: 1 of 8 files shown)


pandapipes/component_models/component_toolbox.py
```diff
@@ -197,4 +197,24 @@ def standard_branch_wo_internals_result_lookup(net):
     else:
         required_results_hyd.extend([("v_mean_m_per_s", "v_mps")])
 
-    return required_results_hyd, required_results_ht
\ No newline at end of file
+    return required_results_hyd, required_results_ht
+
+
+def get_component_array(net, component_name, internal=True):
```
SimonRubenDrauz, line 203: I do not understand, why you call it internal?

pandapipes/component_models/component_toolbox.py
```diff
@@ -197,4 +197,24 @@ def standard_branch_wo_internals_result_lookup(net):
...
+    :param internal: If True, only return entries of active elements (included in _active_pit)
+    :type internal: bool
+    :return: component_array - internal array of the component
+    :rtype: numpy.ndarray
+    """
+    f_all, t_all = get_lookup(net, "branch", "from_to")[component_name]
+    if not internal:
+        return net["_pit"]["components"][component_name]
+    in_service_elm = get_lookup(net, "branch", "active_hydraulics")[f_all:t_all]
+    return net["_pit"]["components"][component_name][in_service_elm]
```
SimonRubenDrauz, line 220: Could it happen, that you get problems with the connectivity check? If you set some components out of service within the connectivtiy check this is only saved in _active_pit, right?

pandapipes/component_models/component_toolbox.py
dlohmeier: Thought it was called internal_pit, but maybe we should rename it to only_active.

pandapipes/component_models/component_toolbox.py
SimonRubenDrauz: Perfect. Thanks for the clarification.

dlohmeier: Actually, this function is just a helper function to retrieve a component pit and save a few duplicated lines of code within the components themselves. But I just realize, that this is not general enough, I need to still distinguish between branch and node components.

dlohmeier: I updated this behavior now.

pandapipes/component_models/component_toolbox.py
dlohmeier: I updated this behavior now.

SimonRubenDrauz: Perfect

pandapipes/component_models/component_toolbox.py
SimonRubenDrauz: Great :)

SimonRubenDrauz (approved):

PR #563, opened 2023-09-08
calling 'calculate_derivatives_thermal' once

SimonRubenDrauz:
- Just calling 'calculate_derivatives_thermal' once instead of calling it for each component seperately
- renaming TINIT and T_OUT in TINIT and  TOUTINIT in branch


pandapipes/pipeflow.py
```diff
@@ -269,17 +269,22 @@ def solve_temperature(net):
     branch_pit[mask, TO_NODE_T] = branch_pit[mask, FROM_NODE]
 
     for comp in net['component_list']:
-        comp.calculate_derivatives_thermal(net, branch_pit, node_pit, branch_lookups, options)
+        comp.adaption_before_derivatives_thermal(
+            net, branch_pit, node_pit, branch_lookups, options)
+    calculate_derivatives_thermal(net, branch_pit, node_pit, options)
+    for comp in net['component_list']:
+        comp.adaption_after_derivatives_thermal(
+            net, branch_pit, node_pit, branch_lookups, options)
```
EPrade, line 277: Just for my understanding: You basically applied the same process as in calculate_hydraulics here right? Does the reduction of the function call to once instead of iterating through components improve calculation speed or is it just for clarity and more organised structure?

pandapipes/component_models/abstract_models/branch_w_internals_models.py
```diff
@@ -6,7 +6,7 @@
 
 from pandapipes.component_models.abstract_models.branch_models import BranchComponent
 from pandapipes.component_models.component_toolbox import set_entry_check_repeat
-from pandapipes.idx_branch import ACTIVE, FROM_NODE, TO_NODE, TINIT, RHO, ETA, CP, ELEMENT_IDX
+from pandapipes.idx_branch import ACTIVE, FROM_NODE, TO_NODE, TINIT_IN, RHO, ETA, CP, ELEMENT_IDX
```
dlohmeier, line 9: What is this renaming of TINIT meant for? We initialize this variable as (TINIT[from_nodes] + TINIT[to_nodes]) / 2. That would mean to me that it is not the temperature at the inlet.

dlohmeier (commented):
In general, I like this whole approach and it's about time to put together all branches when calculating the thermal derivatives. One thing that strikes me is the variable TINIT / TINIT_IN that we define in branch_idx and that you rename here. This is not a solver variable, so I would suggest to throw out this variable fully. It leads to confusion, because "the branch temperature" is not defined. It would make more sense to me to always define which temperatures precisely I need in what constellation. If I want the intermediate of the two neighboring junctions, I have to get from_nodes and to_nodes beforehand.

pandapipes/pipeflow.py
SimonRubenDrauz: I would say it does both. By applying the same procedure as in calculate_hydraulics you increase the readability. But it also increases the performance, by just calling this function once instead of calling it in each component seperately. The reason we can do so is, that the general function remains the same. Component specific changes can be done in ..._before_... and ..._after_... but they should be minor, that's why this is a performance boost.

SimonRubenDrauz:
@dlohmeier: What I do not understand: What do you mean by throwing out the variable? Shall we rename it? I wouldn't throw it out completely because we use it in different functions.

dlohmeier:
Yes, it is used, but it can be replaced by a call like

    from_nodes = branch_pit[:, FROM_NODE].astype(np.int32)
    to_nodes = branch_pit[:, TO_NODE].astype(np.int32)
    t_init = (node_pit[from_nodes, T_INIT] + node_pit[to_nodes, T_INIT]) / 2

The advantage is that there is more control over what value really is needed in the specific case. If we assume that a fluid property of a branch element depends on the branch temperature, it makes sense to rethink how the temperature of this branch should be calculated. In case of thermal calculation, we would definitely need to change the above statement to:

    from_nodes = branch_pit[:, FROM_NODE].astype(np.int32)
    t_init = (node_pit[from_nodes, T_INIT] + branch_pit[:, T_OUT]) / 2

but in case of hydraulic simulation, T_OUT might not be properly initialized. We have to check this again, but I think that it is more useful to assume that the temperature is a very local (point) variable that cannot be a branch variable and each time you need "the temperature" of a branch, you have to define it yourself.

dlohmeier (approved):
This now looks good to me.

SimonRubenDrauz:
@dlohmeier: I just checked how T_OUT is initialized in a pipe. Actually it is exactly what you described above:

grafik

Thus, both equations above should get the same result in the hyadraulics calculation. So we could use this definition:

t_init = (node_pit[from_nodes, T_INIT] + branch_pit[:, T_OUT]) / 2

to always calculate the correct temperature:

SimonRubenDrauz:
I rethought your point and decided to throw out TINIT completely. I agree that it is really misleading and causes confusion. And the result is that I found something what really confuses me. 
In some files we use the branch temperature combined with p_from/p_to. You can find this in the pump_component, result_extraction for example where we determine v_mean/v_from/v_to. I also apated it. I guess, it was a mistake. What do you think?

pandapipes/component_models/pump_component.py
```diff
@@ -102,11 +102,12 @@ def adaption_before_derivatives_hydraulic(cls, net, branch_pit, node_pit, idx_lo
...
-        numerator = NORMAL_PRESSURE * pump_branch_pit[:, TINIT]
+        t_from = node_pit[from_nodes, TINIT_NODE]
+        numerator_from = NORMAL_PRESSURE * t_from
         v_mps = pump_branch_pit[:, VINIT]
         if fluid.is_gas:
             # consider volume flow at inlet
-            normfactor_from = numerator * fluid.get_property("compressibility", p_from) \
+            normfactor_from = numerator_from * fluid.get_property("compressibility", p_from) \
                               / (p_from * NORMAL_TEMPERATURE)
             v_mean = v_mps * normfactor_from
```
dlohmeier, line 112: I think we should go through the whole process again and check out how we get to the norm state (what temperature and pressure do we use?) and then how we can get back from there. Maybe v_mean also should be the mid between v_from and v_to.

pandapipes/test/pipeflow_internals/test_pipeflow_modes.py
```diff
@@ -95,7 +95,7 @@ def test_heat_only(use_numba):
                         nonlinear_method="automatic", mode="hydraulics", use_numba=use_numba)
 
     p = ntw._pit["node"][:, 5]
-    v = ntw._pit["branch"][:, 12]
```
dlohmeier, line 98: Wouldn't it make more sense to import this? Less error prone...

dlohmeier:
I like the adjustments here. Guess we have to look at the pump again, but it seems that you were able to fix a few error prone implementations here.

pandapipes/component_models/pump_component.py
dlohmeier: If we rename v_mean to v_from, it should be fine.

pandapipes/component_models/abstract_models/branch_w_internals_models.py
SimonRubenDrauz: I changed the name of T_OUT into TOUTINIT to be consistent. TININIT is removed.

SimonRubenDrauz:
@dlohmeier, can we merge this PR. All changes have been realized which you recommended.

dlohmeier (approved):
Looks really good to me. This should improve performance and be a great preparation of using numba for thermal simulation as well, and we got rid of yet another branch variable. :-D

PR #564, opened 2023-09-12
pipeflow cancellation if all nodes are out-of-service after the connectivity check

SimonRubenDrauz:


dlohmeier (commented):
Great to have this covered. I am just wondering, if it would make more sense to raise an Exception instead of just logging a warning and interrupting the pipeflow. What is the idea behind this? And could we also check net.converged?

SimonRubenDrauz:
Right now, what happens is, that it just returns wihtout raising any error. This means that the pipeflow can be conducted even in an empty or non-supplied net. The results would be nans for all nodes, pipes and so on. I prefer this solution as it is consistent if only a certain part of the net is not supplied. 
By default calling the pipeflow, net.converged is set False by default. Shall we set it to True or leave it False? I would add it to the test accordingly!

SimonRubenDrauz:
@dlohmeier: any new thoughts on this?

dlohmeier (dismissed):
All right, I think that's a good solution.

SimonRubenDrauz:
changed fork branch. Please refer to #579

PR #558, opened 2023-08-16
Update .readthedocs.yml

jkisse:
According to their blogpost https://blog.readthedocs.com/use-build-os-config/

PR #565, opened 2023-09-12
readthedocs.yml deprecation

SimonRubenDrauz:
see also [read our blog post](https://blog.readthedocs.com/use-build-os-config/)

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,4 +10,4 @@ python:
        - method: pip
          path: .
          extra_requirements:
-            - docs
+            - doc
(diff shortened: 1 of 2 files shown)


dlohmeier (approved):

jkisse:
fyi, I tried something similar already in #558 and it did not work (RTD build was failing) so I had to revert it. Have you already tested RTD based on your feature branch, @SimonRubenDrauz ?

SimonRubenDrauz:
@jkisse, thanks for your comment. I actually also updated the webhook for RTD. As you can see it is also checking if the RTD doc is correctly built what was successful for this PR. Is this what you mean?

PR #568, opened 2023-09-29
Bugfix/generic geodata

SimonRubenDrauz:
Only pipes were considered so far. Now all Branch components are considered.


pandapipes/plotting/generic_geodata.py
```diff
@@ -50,19 +50,20 @@ def build_igraph_from_ppipes(net, junctions=None):
...
     pp_junction_mapping = dict(list(zip(junction_index, list(range(nr_junctions)))))
 
-    mask = _get_element_mask_from_nodes(net, "pipe", ["from_junction", "to_junction"], junctions)
-    for pipe in net.pipe[mask].itertuples():
-        g.add_edge(pp_junction_mapping[pipe.from_junction], pp_junction_mapping[pipe.to_junction],
-                   weight=pipe.length_km)
+    #mask = _get_element_mask_from_nodes(net, "pipe", ["from_junction", "to_junction"], junctions)
+    #for pipe in net.pipe[mask].itertuples():
+    #    g.add_edge(pp_junction_mapping[pipe.from_junction], pp_junction_mapping[pipe.to_junction],
+    #               weight=pipe.length_km)
```
dlohmeier, line 56: If this part is not needed, I think we should just delete it.

dlohmeier (dismissed):
Thanks for the adjustment! Just one small comment added.

PR #569, opened 2023-10-05
Added pipe standard type 'heating pipe' 

EPrade:
Added a standard type for heating pipes.

The reason for this addition is, that in literature and industry the parameter for heat loss specified is predominantly U [W/mK]  instead of alpha [W/m²K]. 

- I added a csv file from a DH pipe manufacturer for parameters ajar to the existing pipe.csv.
- In create_empty_network a new function parameter 'type' has been added to chose between 'generic', 'gas' and 'heating'. At the moment 'heating' uses the new data and 'generic' and 'gas' use the old. Maybe we should specify here for water(if there are specific parameters for this type of pipes?) or just drop either 'gas' or 'generic' and only differentiate between gas and not gas.
- I created the function 'add_heating_std_types' from 'add_basic_std_types' which extracts data from the new csv and calculates alpha from the given U values


SimonRubenDrauz:
Thanks for the contribution. I think this is a good idea. In my opinion, however, I think we should not add a second pipe.csv. I would rather extend the exisiting pipe.csv and add Nones for the pipes which do not have U values. The reason is that, if you implement it your way, at two different places the user needs to decide whether she/he has a heating or a non-heating system - at the beginning, when he/she creates the net, and during the pipeflow (this could lead to unitended errors). Furthermore, in future it could be that it might be interesting to investigate a heating/gas grid. For example in case of a steam network. In that case we would need to change your implementation again, which would affect convert_format and make thinks quite comlicated. This is my feeling. @dlohmeier, what do you think?

EPrade:
@SimonRubenDrauz @dlohmeier I updated the pull request. Could you take a look again?

pandapipes/create.py
```diff
@@ -429,13 +430,23 @@ def create_pipe(net, from_junction, to_junction, std_type, length_km, k_mm=0.2,
...
+    from pandapipes.toolbox import calculate_alpha
     add_new_component(net, Pipe)
 
     index = _get_index_with_check(net, "pipe", index)
     _check_branch(net, "Pipe", index, from_junction, to_junction)
     _check_std_type(net, std_type, "pipe", "create_pipe")
 
     pipe_parameter = load_std_type(net, std_type, "pipe")
+
+    if pd.notna(pipe_parameter['u_w/mk']) and alpha_w_per_m2k == 0.0:
```
dlohmeier, line 442: I would prefer to use u_w_per_mk here in analogy to the naming of other parameters.

pandapipes/std_types/std_types.py
```diff
@@ -5,7 +5,7 @@
 import os
 import warnings
 import re
-
+import numpy as np
```
dlohmeier, line 8: I think this import statement is not needed, would look better wrt codacy to remove.

pandapipes/toolbox.py
```diff
@@ -615,3 +615,6 @@ def get_internal_tables_pandas(net, convert_types=True):
                         tbl[col] = tbl[col].astype(np.bool_)
 
     return node_table, branch_table
+
+def calculate_alpha(d, u):
```
dlohmeier, line 619: I would rename this function and signature: def alpha_from_u(u, s) . As far as I can recall using s for the wall thickness is more common than d, which is usually the diameter. Otherwise, it could make sense to name it precisely as thickness.

dlohmeier: And to avoid confusion, I would write some docs, describing the meaning of alpha, u and s.

pandapipes/std_types/library/Pipe.csv
```diff
@@ -1,221 +1,239 @@
...
-700_ST<16;700;711.0;696.8;100.14;ST
-800_ST<16;800;813.0;797.0;101.62;ST
-900_ST<16;900;914.0;894.0;91.4;ST
-1000_ST<16;1000;1016.0;996.0;101.6;ST
-1200_ST<16;1200;1220.0;1198.0;110.91;ST
-1400_ST<16;1400;1420.0;1398.0;129.09;ST
-1600_ST<16;1600;1620.0;1588.0;101.25;ST
-1800_ST<16;1800;1820.0;1785.0;104.0;ST
-2000_ST<16;2000;2020.0;1980.0;101.0;ST
+std_type;nominal_width_mm;outer_diameter_mm;inner_diameter_mm;standard_dimension_ratio;material;u_w/mk
```
dlohmeier, line 1: u_w/mk to u_m_per_m2k

pandapipes/create.py
dlohmeier: when renaming u to alpha in table, this is not needed anymore

PR #571, opened 2023-11-03
made list of liquids and gases global in fluids.py

dlohmeier:
A first step towards an applicable fluid library, if needed. Anyone from outside at least can get a better overview of the implemented fluids in pandapipes.


dlohmeier:
@kbensafta could you provide a lhv for the new biomethane properties? You can just write them here, I could add them.

SimonRubenDrauz (approved):

pandapipes/properties/fluids.py
```diff
@@ -668,36 +671,31 @@ def linear_property(prop):
...
+    properties["molar_mass"] = constant_property("molar_mass")
+    properties["der_compressibility"] = constant_property("der_compressibility")
+    properties["compressibility"] = linear_property("compressibility")
+
+    if phase == 'gas':
+        for entry, name in [("lhv", "lower_heating_value"), ("hhv", "higher_heating_value")]:
+            try:
+                properties[entry] = constant_property(name)
+            except FileNotFoundError:
+                pass
```
SimonRubenDrauz, line 696: I would add a Warning here.

SimonRubenDrauz (dismissed):
I really like the implementation. I just would add a Warning in case of not finding lhv/hhv!

pandapipes/properties/fluids.py
dlohmeier: warning was added.

PR #592, opened 2024-01-16
Added unsupplied_junctions and elements_on_path to pandapipes

mfranz13:
Added unsupplied_junctions and elements_on_path with tests to pandapipes.
Added the convenience functions get_all_branch_component_table_names and get_all_branch_component_models to prevent hardcoding the branch element table names like in pandapower.

### DIFF
--- a/src/pandapipes/topology/graph_searches.py
+++ b/src/pandapipes/topology/graph_searches.py
@@ -186,9 +186,3 @@ def elements_on_path(mg, path, element="pipe", check_element_validity=True):
     else:
         return [mg.get_edge_data(b1, b2)["key"][1] for b1, b2 in zip(path, path[1:])
                 if mg.get_edge_data(b1, b2)["key"][0] == element]
-
-
-if __name__ == '__main__':
-    import pandapipes.networks as nw
-    net = nw.gas_meshed_delta()
-    dist = calc_minimum_distance_to_junctions(net, net.ext_grid.junction.values)
--- a/src/pandapipes/topology/graph_searches.py
+++ b/src/pandapipes/topology/graph_searches.py
@@ -117,7 +117,7 @@ def unsupplied_junctions(net, mg=None, slacks=None, respect_valves=True):
     return not_supplied
 
 
-def elements_on_path(mg, path, element="pipe"):
+def elements_on_path(mg, path, element="pipe", check_element_validity=True):
     """
      Finds all elements that connect a given path of junctions.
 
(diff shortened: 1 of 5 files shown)


src/pandapipes/topology/graph_searches.py
```diff
@@ -78,6 +79,79 @@ def calc_minimum_distance_to_junctions(net, junctions, notravjunctions=None, nog
...
+     Finds all elements that connect a given path of junctions.
+
+     INPUT:
+        **mg** (NetworkX graph) - NetworkX Graph or MultiGraph that represents a pandapipes network.
+
+        **path** (list) - List of connected junctions.
+
+        **element** (string, "l") - element type of type BranchComponent
+
+        **multi** (boolean, True) - True: Applied on a NetworkX MultiGraph
```
dlohmeier, line 131: I guess this one is not needed?

src/pandapipes/topology/graph_searches.py
```diff
@@ -78,6 +79,79 @@ def calc_minimum_distance_to_junctions(net, junctions, notravjunctions=None, nog
...
+     EXAMPLE:
+         import topology as top
+
+         mg = top.create_nxgraph(net)
+         elements = top.elements_on_path(mg, [4, 5, 6])
+
+     """
+    table_names = get_all_branch_component_table_names()
+    if element not in table_names:
+        raise ValueError("Invalid element type %s" % element)
```
dlohmeier, line 146: As far as I understand, this check is not required to prevent errors. Could we make it optional via another default argument?

src/pandapipes/topology/topology_toolbox.py
```diff
@@ -0,0 +1,42 @@
...
+def get_all_branch_component_models():
+    """
+    Get all models of available branch components
+
+    :return: branch model
+    :rtype: list
+    """
+    def get_all_subclasses(cls):
+        all_subclasses = list()
+        for subclass in cls.__subclasses__():
```
dlohmeier, line 18: Does `__subclasses__` search for appearances in the whole python path? If not, we might need to find a different implementation that checks the appearing classes inside the net...

dlohmeier (commented):
Thanks a lot for the great additional features! I like this implementation. just added a few small comments.

src/pandapipes/topology/topology_toolbox.py
mfranz13: Yes, __subclasses__ includes all child classes and is not dependent from net

mfranz13:
@dlohmeier thx for your suggestions, i implemented the changes :)

src/pandapipes/topology/graph_searches.py
```diff
@@ -78,6 +79,115 @@ def calc_minimum_distance_to_junctions(net, junctions, notravjunctions=None, nog
...
+            raise ValueError("Invalid element type %s" % element)
+    if isinstance(mg, nx.MultiGraph):
+        return [edge[1] for b1, b2 in zip(path, path[1:]) for edge in mg.get_edge_data(b1, b2).keys()
+                if edge[0] == element]
+    else:
+        return [mg.get_edge_data(b1, b2)["key"][1] for b1, b2 in zip(path, path[1:])
+                if mg.get_edge_data(b1, b2)["key"][0] == element]
+
+
 if __name__ == '__main__':
```
dlohmeier, line 191: I think this is not really necessary, the file should just be seen as a module from where to load functions. Can we just delete the "main" part?

dlohmeier (approved):
Looks good to me. Thanks a lot for this great added feature!

PR #596, opened 2024-02-23
Props at norm state

SimonRubenDrauz:
bugfix: density needs to be determined under norm state not under the given temperatures


SimonRubenDrauz:
update develop merge in a different branch

PR #598, opened 2024-02-28
renamed JAC_DERIV_DT1 to JAC_DERIV_DTOUT in branch_idx

dlohmeier:
I think using TOUT here acknowledges that we are always using the pipe outlet temperature considering the flow direction, which is also been reflected by the naming of the corresponding variable TOUTINTIT.


dlohmeier:
We might also change JAC_DERIV_DT to JAC_DERIV_DTIN, but this is not reflected by any other variable, as the temperature used here belongs to the node (so it could be called JAC_DERIV_DTNODE as well).

PR #602, opened 2024-04-04
Made some node entries more global than just in pipe

dlohmeier:
I shifted some node entries from pipe component to branch_w_internals_models.py


SimonRubenDrauz:
Looks good!

SimonRubenDrauz (approved):

PR #612, opened 2024-04-15
Changed  heat coefficient from alpha to u

EPrade:
changed "alpha_w_per_m2k" to "u_w_per_m2k":

- changed tutorials
- added district heating standard pipes
- added"u_w_per_m2k" in std types for pipes
- added deprecation warning for alpha

Not sure how to handle the version and format version. Since i changed the format I need to change the format version. Do I also need to change the "normal" version? After the pull request develop would then be on a new version


src/pandapipes/create.py
```diff
@@ -439,8 +437,8 @@ def create_pipe(net, from_junction, to_junction, std_type, length_km, k_mm=0.2,
     pipe_parameter = load_std_type(net, std_type, "pipe")
```
dlohmeier, line 437: Could we add a warning here, if "alpha_w_per_m2k" is contained in the kwargs? Just to make the change more transparent.

src/pandapipes/create.py
```diff
@@ -501,20 +498,29 @@ def create_pipe_from_parameters(net, from_junction, to_junction, length_km, diam
...
     _check_branch(net, "Pipe", index, from_junction, to_junction)
 
+
+    #Check if alpha parameter was declared (Alpha has been changed to U)
+    if 'alpha_w_per_m2k' in kwargs:
+
+        if u_w_per_m2k == 0:
+            u_w_per_m2k = kwargs['alpha_w_per_m2k']
+
+        raise DeprecationWarning(' In the future alpha will be renamed to u')
```
dlohmeier, line 516: Raising a deprecation warning interrupts the code flow. What we need here is just an output. I would preferrably stick with the python standard of using the warnings module for this, instead of our normal logger. So what would be done here is:

    import warnings
    warnings.warn("The parameter alpha_w_per_m2k has been renamed to u_w_per_m2k."
                       "It will be removed in future.", DeprecationWarning)

Preferrably, we'd also like to make sure with some tests, that the behavior is as expected. This can be done with

    import pytest
    with pytest.warns(DeprecationWarning):
        pandapipes.create_pipe...(with alpha)
        assert ... # make sure that u is the given alpha value

src/pandapipes/create.py
```diff
@@ -1439,11 +1445,26 @@ def create_pipes_from_parameters(net, from_junctions, to_junctions, length_km, d
     index = _get_multiple_index_with_check(net, "pipe", index, len(from_junctions))
     _check_branches(net, from_junctions, to_junctions, "pipe")
 
+    if 'alpha_w_per_m2k' in kwargs:
+
+        if u_w_per_m2k == 0:
+            u_w_per_m2k = kwargs['alpha_w_per_m2k']
+
+        raise DeprecationWarning(' In the future alpha will be renamed to u')
```
dlohmeier, line 1453: Here the same applies as for the comment above. The equivalence must be tested in a vectorized form. That is:

    u_zero = u_w_per_m2k == 0
    if np.any(u_zero):
        alpha = kwargs['alpha_w_per_m2k']
        if hasattr(alpha, "len"):
            u_w_per_m2k[u_zero] = np.array(list(alpha))[u_zero]
       else:
            u_w_per_m2k[u_zero] = alpha
       # now warn

I guess this should work.

src/pandapipes/io/convert_format.py
```diff
@@ -45,6 +45,9 @@ def _rename_columns(net):
             else:
                 net['controller'].drop('controller', inplace=True, axis=1)
         net["controller"].rename(columns={"controller": "object"}, inplace=True)
+    if "pipe" in net:
+        if "u_w_per_m2k" not in net["pipe"]:
```
dlohmeier, line 49: I think there are some caveats here wrt. where pandas looks for u. Rather do:

    if "u_w_per_m2k" not in net["pipe"].columns

src/pandapipes/create.py
EPrade: @dlohmeier I added two tests. What is your opinion?

dlohmeier (approved):
Looks good, thanks!

PR #614, opened 2024-04-17
Fix/pressure plot

mfranz13:
- circulation pumps are considered as additional deafult slack nodes
- unsupplied nodes are not considered in the plot and raise a user warning


EPrade (commented):
Looks good. Thanks for the fix

EPrade (approved):

PR #615, opened 2024-04-18
introducing bidirectional calculation

SimonRubenDrauz:


src/pandapipes/pipeflow.py
```diff
@@ -68,136 +67,147 @@ def pipeflow(net, sol_vec=None, **kwargs):
...
-        reduce_pit(net, node_pit, branch_pit, mode="heat_transfer")
-        heat_transfer(net)
-        if not net.converged:
-            raise PipeflowNotConverged("The heat transfer calculation did not converge to a "
-                                       "solution.")
-        extract_results_active_pit(net, mode="heat_transfer")
-    elif not calculate_hydraulics:
+    identify_active_nodes_branches(net)
+
+    if calculation_mode == 'heat': use_given_hydraulic_results(net, sol_vec)
```
dlohmeier, line 81: I would like to avoid one-line if conditions. They are not marked by codacy, but I find it less readable.

src/pandapipes/pipeflow.py
```diff
@@ -296,67 +304,50 @@ def set_damping_factor(net, niter, error):
...
-        logger.debug("errorp: %s" % error_2[niter])
-        logger.debug("alpha: %s" % get_net_option(net, "alpha"))
-    else:
-        logger.debug("errorT: %s" % error_1[niter])
-        logger.debug("alpha: %s" % get_net_option(net, "alpha"))
+    for error, var, tol in zip(errors.values(), vars, tols):
+        converged = error[niter] <= tol
+        if not converged: break
+        logger.debug("error_%s: %s" % (var, error[niter]))
+    net.converged = converged and residual_norm <= tol_res
```
dlohmeier, line 340: I am not quite getting how converged is defined here... the value seems not to be set in most cases.

dlohmeier, line 340: Should we discuss the value of tol_res? It is by default 1e-3 and it seems that it is hardly changed in any real world example. But does that value make sense? Is it realistic for all different types of examples and for each of the variables to be solved for? Mass flow can be extremely small in some cases, does 1e-3 make sense there?

src/pandapipes/pipeflow.py
```diff
@@ -296,67 +304,50 @@ def set_damping_factor(net, niter, error):
...
+        logger.debug("alpha: %s" % get_net_option(net, "alpha"))
+        for error_increased, var, val, pit in zip(errors_increased, vars, vals_old, pit_names):
+            if error_increased:
+                net["_active_pit"][pit][:, globals()[var.upper() + 'INIT']] = val
+        if get_net_option(net, "alpha") != 1:
+            net.converged = False
+            return
+    elif nonlinear_method == "std":
+        max_std = np.std(list(diff.values()), axis=1)[:5].max()
+        alpha = 0.33 if max_std == 0 else 1 if max_std < 1 else 0.33 if max_std > 3 else 1 / max_std
```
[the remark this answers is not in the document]
SimonRubenDrauz: Unfortunately I realized it is not working as intended. Therefore, I removed it.

dlohmeier:
Was replaced by #622

PR #618, opened 2024-04-28
Stanet converter 2.0

SimonRubenDrauz:


src/pandapipes/converter/stanet/table_creation.py
```diff
@@ -377,8 +389,8 @@ def create_control_components(net, stored_data, index_mapping, net_params, add_l
             stanet_is_closed=fully_closed[is_pc],
             stanet_flow_kgps=flow[is_pc],
             stanet_active=control_table.ISACTIVE.values[is_pc].astype(np.bool_),
-            **add_info
-        )
+            max_mdot_kg_per_s=control_table.QSOLL.values[is_pc].astype(float) / 3600 * fluid.get_density(NORMAL_TEMPERATURE),
+            **add_info)
```
dlohmeier, line 393: For this, we need #619 to be merged? Is max_mdot_kg_per_s then always required? Can we add the column to the press_control as it it implemented currently?

dlohmeier (commented):
In general, this all looks good to me and should add some clarification wrt. dtypes. Could you just check if the pressure control implementation works as intended?

PR #635, opened 2024-06-20
Feature/circ pump branch model

SimonRubenDrauz:
forbidding mass to disappear/appear


src/pandapipes/idx_node.py
```diff
@@ -23,5 +22,9 @@
 NODE_TYPE_T = 10
 EXT_GRID_OCCURENCE = 11
 EXT_GRID_OCCURENCE_T = 12
+CIRC_PUMP_OCCURENCE = 13
```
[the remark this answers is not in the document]
SimonRubenDrauz: I guess this is not working, as it needs to work for any kind of circ pump. Imagine you have a circ_pump_mass and a circ_pump_pressure in combination at the same junction, then you need the information component-wide.

src/pandapipes/component_models/abstract_models/circulation_pump.py
```diff
@@ -34,13 +33,23 @@ def get_component_input(cls):
...
+        Gets the result table.
+
         :param net: The pandapipes network
         :type net: pandapipesNet
         :return: (columns, all_float) - the column names and whether they are all float type. Only
                 if False, returns columns as tuples also specifying the dtypes
         :rtype: (list, bool)
         """
-        return ["mdot_flow_kg_per_s", "deltap_bar"], True
+        if get_fluid(net).is_gas:
```
dlohmeier, line 44: Can we generally remove velocity results here? Would make much more sense, since we don't need a diameter for simulation anymore.

src/pandapipes/component_models/component_toolbox.py
```diff
@@ -159,7 +159,10 @@ def set_fixed_node_entries(net, node_pit, junctions, eg_types, p_values, t_value
                                     + press_sum) / (number + node_pit[index, eg_count_col])
         node_pit[index, type_col] = typ
         node_pit[index, eg_count_col] += number
-
+        if eg_type == 'p':
```
dlohmeier, line 162: somehow this part seems a bit lost here. I can't really find a perfect spot for it, but this does quite a bit of magic, which I am a bit sceptical of. It also means that some branch components will overwrite this value, if not well taken care of. Of course this happens with all entries touched in this function, but somehow I find hard to predict if this is completely safe to use with just some default value.

src/pandapipes/idx_branch.py
```diff
@@ -30,13 +31,16 @@
...
 TOUTINIT = 23  # Internal slot for outlet pipe temperature
-JAC_DERIV_DT_NODE = 24  # Slot for the derivative fpr T for the nodes connected to branch
-LOAD_VEC_NODES_T = 25 # Slot for the load vector of the nodes connected to branch
-FROM_NODE_T_SWITCHED = 26 # flag to indicate if the from and to node are switched in the thermal calculation
-QEXT = 27  # heat input into the branch [W]
-TEXT = 28 # temperature of surrounding [K]
-PL = 29 # Pressure lift [bar]
-TL = 30 # Temperature lift [K]
-BRANCH_TYPE = 31  # branch type relevant for the pressure controller
+JAC_DERIV_DT_NODE_B = 24  # Slot for the node equation derivative of T for the corresponding branch
```
dlohmeier, line 34: What is the difference between DT_NODE_B and DTOUT?

src/pandapipes/io/file_io.py
```diff
@@ -150,7 +150,8 @@ def from_json_string(json_string, convert=False, encryption_key=None):
     if encryption_key is not None:
         json_string = decrypt_string(json_string, encryption_key)
 
-    net = json.loads(json_string, cls=PPJSONDecoder, registry_class=FromSerializableRegistryPpipe)
+    net = json.loads(json_string, cls=PPJSONDecoder,
+                     object_hook=partial(pp_hook, registry_class=FromSerializableRegistryPpipe))
```
dlohmeier, line 154: I am wondering why a change is necessary here... is it due to a new version in pandapower? Should we add it to a different PR?

src/pandapipes/pf/build_system_matrix.py
```diff
@@ -32,145 +33,201 @@ def build_system_matrix(net, branch_pit, node_pit, heat_mode):
...
+            system_rows[len_pc:len_slack] = slack_nodes
+
+            # mass flow slack equation
+            # --------------
+            # from_slack_dF_dm
+            system_cols[len_slack:len_fsb] = branch_matrix_indices[slack_branches_from]
+            system_rows[len_slack:len_fsb] = slack_mass_matrix_indices[slack_masses_from]
+            # to_slack_dF_dm
+            system_cols[len_fsb:len_tsb] = branch_matrix_indices[slack_branches_to]
+            system_rows[len_fsb:len_tsb] = slack_mass_matrix_indices[slack_masses_to]
```
dlohmeier, line 182: duplicated code

src/pandapipes/pf/derivative_calculation.py
```diff
@@ -86,23 +88,36 @@ def calculate_derivatives_thermal(net, branch_pit, node_pit, _):
...
+    branch_pit[:, LOAD_VEC_BRANCHES_T] = cp * m_init * (-t_init_i + t_init_i1 - tl) - alpha * (
+                t_amb - t_m) * length + qext
 
     branch_pit[:, JAC_DERIV_DT] = - cp * m_init + alpha / 2 * length
     branch_pit[:, JAC_DERIV_DTOUT] = cp * m_init + alpha / 2 * length
 
-    branch_pit[:, JAC_DERIV_DT_NODE] = m_init
-    branch_pit[:, LOAD_VEC_NODES_T] = m_init * t_init_i1
+    branch_pit[:, JAC_DERIV_DT_NODE_B] = m_init * cp_out
+    branch_pit[:, JAC_DERIV_DT_NODE_N] = - m_init * cp_n
```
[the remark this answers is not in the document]
SimonRubenDrauz: This is the correct derivative. Before did not change cp, we said, it is constant. However, this is not correct. DT_NODE_B is related to TOUT of the branch, while DT_NODE_N is related to the mixed temperature of each node TINIT.

src/pandapipes/io/file_io.py
SimonRubenDrauz: I changed it back. No clue, where it came from. Not required.

src/pandapipes/pf/derivative_calculation.py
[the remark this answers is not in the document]
dlohmeier: I still don't get this point. I thought that what you do is sum up the power flowing into the node with SUM(mi * cpi * touti) and the power leaving the node and thus entering the branch with SUM(mj * cp_n * t), which means that we have to calculate JAC_DERIV_DT_NODE_N with a relation to the from_node and then sum up this value for all branches connected to the respective node on the from-side (i.e. leaving the node).

src/pandapipes/component_models/abstract_models/circulation_pump.py
SimonRubenDrauz: Removed them in most wo internal components except from valves. Velocities in valves seem to be relevant as most tools keep this information.

src/pandapipes/component_models/component_toolbox.py
SimonRubenDrauz: I made it a bit more explicit and directly connected to each component!

PR #636, opened 2024-06-26
-fixed heat consumer result extraction

EPrade:


dlohmeier (commented):
Thanks for the correction. I guess you had problems with the previous implementation, so could you provide a short example and add it to the tests, such that the test fails with the previous version and passes with the change?

EPrade:
> Thanks for the correction. I guess you had problems with the previous implementation, so could you provide a short example and add it to the tests, such that the test fails with the previous version and passes with the change?

@dlohmeier I added a test. Do you think its sufficient?

dlohmeier (approved):
Perfect, thanks a lot!

PR #640, opened 2024-07-26
Bugfix: result extraction for incompressible media now based on real density

dlohmeier:


EPrade (approved):

SimonRubenDrauz:
I see your point but I don't agree totally with your adaptions. vf is actually defined as standard flow rate. This is also extracted as result in liquid fluids. If you change vf the way you do, the results are misleading in my opinion. I guess we have to possibilities:
1.) Either we use vf differently. In liquids it is considered as the volumetric flow rate (and displayed accordingly), in gases it is the standard flow rate
2.) We consider both in the results and display both. 

In terms of the velocities I agree and we should correct the bug accordingly

SimonRubenDrauz (approved):

dlohmeier:
pandapower only supports python >=3.9 on develop. The tutorial tests are still running with pandapower develop, which is why also required tests fail, not just the relying tests. I guess we should change this to pandapower master? Should I do it in this PR? @SimonRubenDrauz @EPrade

src/pandapipes/converter/stanet/preparing_steps.py
EPrade, line 1: What is the reason for this change?

dlohmeier: Ah, sorry, my bad, should not belong to this PR.

dlohmeier:
> I see your point but I don't agree totally with your adaptions. vf is actually defined as standard flow rate. This is also extracted as result in liquid fluids. If you change vf the way you do, the results are misleading in my opinion. I guess we have to possibilities: 1.) Either we use vf differently. In liquids it is considered as the volumetric flow rate (and displayed accordingly), in gases it is the standard flow rate 2.) We consider both in the results and display both.
> 
> In terms of the velocities I agree and we should correct the bug accordingly

This is now handled differently for compressible / incompressible media

SimonRubenDrauz (approved):

PR #648, opened 2024-08-25
t_outlet in pipe results

EPrade:


src/pandapipes/pf/result_extraction.py
```diff
@@ -62,7 +62,8 @@ def get_basic_branch_results(net, branch_pit, node_pit):
     t1 = node_pit[to_nodes, TINIT_NODE]
     vf = branch_pit[:, MDOTINIT] / get_fluid(net).get_density(NORMAL_TEMPERATURE)
     v = branch_pit[:, MDOTINIT] / fluid.get_density(NORMAL_TEMPERATURE) / branch_pit[:, AREA]
-    return v, branch_pit[:, MDOTINIT], vf, from_nodes, to_nodes, t0, t1, branch_pit[:, RE], \
+    t_outlet = branch_pit[:, TOUTINIT]
```
dlohmeier, line 65: Since there are so many different values extracted here, we might just move the dictionary creation to this function, instead of returning so many values and then just making a dictionary in the calling function.

src/pandapipes/pf/result_extraction.py
```diff
@@ -212,6 +213,11 @@ def extract_branch_results_with_internals(net, branch_results, table_name,
 
             for i, (res_name, entry) in enumerate(res_mean_hydraulics):
                 res_table[res_name].values[pt] = res[i + 3][connected_ind] / num_internals
+    #outlet temperature
+    use_numba = get_net_option(net, "use_numba")
+    indices, sections = _sum_by_group(use_numba, branch_pit[:, ELEMENT_IDX], np.ones(len(branch_pit)))
+    indices_last_section = (np.cumsum(sections)-1).astype(int)
+    res_table["t_outlet_k"] = branch_results["t_outlet"][indices_last_section]
```
dlohmeier, line 220: I can imagine that there are still some tests failing that check indexing. Please have another look at how this is done in other sections. You need to consider the connected_thermal lookup and you should work on the underlying numpy values of the result table and need to match the pit indices to the result table indices.

src/pandapipes/pf/result_extraction.py
```diff
@@ -212,6 +213,17 @@ def extract_branch_results_with_internals(net, branch_results, table_name,
 
             for i, (res_name, entry) in enumerate(res_mean_hydraulics):
                 res_table[res_name].values[pt] = res[i + 3][connected_ind] / num_internals
+        if len(res_branch) > 0:
+            use_numba = get_net_option(net, "use_numba")
+            indices, sections, connected_sum = _sum_by_group(use_numba, idx_pit, np.ones_like(idx_pit),
```
dlohmeier, line 218: Interesting enough, we do not need indices here anymore. You could replace it by _

dlohmeier (commented):
Looks good to me. Great addition for result handling. Could you add a change to the log?

dlohmeier (approved):
Perfect, thanks!

dlohmeier (approved):

PR #651, opened 2024-09-26
Edit default output writer

harrahx3:
Remove the results from source, sink and ext_grid from the default OutputWriter and edit the testing accordingly, because cannot assume that the network will contain these elements.

I was running a timeseries calculation with a District Heating Network, which has no source, sink and ext_grid and figured out that I need to explicitly defined an OutputWriter as the default one assume that these elements are present in the network.

It seems safe to assume that junctions and pipes are always part of the network though.

### DIFF
--- a/src/pandapipes/test/pipeflow_internals/test_time_series.py
+++ b/src/pandapipes/test/pipeflow_internals/test_time_series.py
@@ -79,6 +79,7 @@ def _data_source():
 
 
 def _compare_results(ow):
+    _compare_results_default_ow(ow)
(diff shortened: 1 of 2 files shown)


dlohmeier (commented):
Thanks for this Pull Request. You are absolutely right that it is not quite logical to assume that a default pandapipes output write must include sinks, sources and ext_grids. I just think that we should do the following as well:
1. Make sure that an output writer neglects logging variables to be added in case the table does not exist (you might add an enforce argument)
2. Small auxiliary functions to add logging variables for gas and heating networks.
Is it possible for you to add to this PR? Otherwise, we have to add this to our schedule.

PR #673, opened 2024-12-05
speeding up tests in GA

jthurner:
This PR cleans up the GA workflow files.

* tests on commits pushed to dev / feature branches now finish in ~4min (was: 13min)
* total CI time for these tests is at ~25min (was: ~2h)

### DIFF
--- a/.github/workflows/relying_test_on_dev_merge.yml
+++ b/.github/workflows/relying_test_on_dev_merge.yml
@@ -1,0 +1,34 @@
+# pandapipes-develop branch is designed to work with pandapower-master branch, but checks the ability to
+# work with pandapower-develop branch (relying tests, usually not required for merges) in order to avoid
+# problems in future releases
+
+name: ppipes_relying
+on:
+  pull_request:
+    types:
+      - closed
+  workflow_dispatch:
+
+jobs:
+  relying:
+    name: Relying tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python-version: ['3.9', '3.10', '3.11', '3.12']
+        os:  [ ubuntu-latest, windows-latest ]
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


jthurner:
I had removed the "pip list" steps on purpose because uv already provides a pretty readable output of installed packaged - added it back in now. 

There are some changes here that seemed reasonable to me but should be reviewed:

### Testing without Numba
Some pytest runs were repeated after uninstalling numba. But the tests already contain parametrization (e.g. [here](https://github.com/retoflow/pandapipes/blob/f5ab7bef2e5b5440eda0a916d2506d2ff83cd89d/src/pandapipes/test/api/test_components/test_flow_control.py#L7-L7)) to run with/without numba enabled, so in effect these tests were run once with numba enabled and 3 times with numba disabled. This seems unnecessary - removed the uninstall-numba-and-retest steps from all but the tutorial tests (as the notebook examples are not parametrized to run without numba).

### relying tests
These run the whole test matrix again against the latest pandapower develop at the time the action runs (or that was the idea, in reality they ran once more against the latest pp release because of the order of the install commands). A comment explains their intention as an early-warning and that they are  "usually not required for merges". 

To not have them run on every single push, I moved them to "on merge". This means that they will only flag a problem after you merged the PR, and only as a notification. They will still run on pushes to master.

Interestingly, pandapower also runs "relying" tests against pandapipes develop, but these have been [failing](https://github.com/e2nIEE/pandapower/actions/runs/12231858517/job/34115730491) for some time. I would have expected them to succeed/fail in sync if run at roughly the same time. pandapower does [some things differently](https://github.com/e2nIEE/pandapower/blob/37db17486f4cd2d1d57a3f88cbab31b5374d0124/.github/workflows/github_test_action.yml#L176-L188), like installing simbench:develop and calling the tests in a a different way.

SimonRubenDrauz (approved):

Issue #679, opened 2024-12-19
Bidirectional calculation failing in timeseries calculation after qext_w was zero in a timestep


JonasPfeiffer123:
Had this issue a few weeks ago and described it in a more or less relevant comment section so it wasn't really visible and I forgot about it. As I'm now focusing on this issue again, I thought I'd open a new issue here. The issue occurs specifically when qext_w is zero. When I set it to 1 for example it works. How am I supposed to handle inactive heat consumers in time steps, how should I handle time steps where all heat consumers could be zero? The following is the description of the previous post.:

Got a "minimal" example running where the same issue occurs. The following code includes all helper functions and classes I'm currently using. In the test function 20 time steps are calculated. When setting one qext_w=0 in the first heat consumer (here in time step 8) The calculation fails in time step 9 with 
```
pandapower\timeseries\run_time_series.py", line 82, in pf_not_converged
    raise ts_variables['errors'][0] pandapipes.pf.pipeflow_setup.PipeflowNotConverged
``` 
The more important Information is coming from the output after time step 8: 
```
"WARNING:pandapipes.component_models.heat_consumer_component:qext_w is equals to zero for heat consumers with index Index([0], dtype='int64'). Therefore, the defined temperature control cannot be maintained.
INFO:pandapipes.pf.pipeflow_setup:Setting the following nodes out of service for heat_transfer calculation in connectivity check:
In table junction: [1, 2, 3, 4, 5, 6, 7]
dgstrf info 8
C:\Users\jonas\AppData\Local\Programs\Python\Python311\Lib\site-packages\pandapipes\pipeflow.py:255: MatrixRankWarning: Matrix is exactly singular
  x = spsolve(jacobian, epsilon)
INFO:pandapipes.pf.pipeflow_setup:Setting the following nodes out of service for heat_transfer calculation in connectivity check:
In table junction: [1, 2, 3, 4, 5, 6, 7]
dgstrf info 8"
```

In my understanding the nodes connected to the first heat_consumer are set out of service as the heat_consumer results in NaN-Values and as this are connection nodes to all other nodes this sets all other nodes out of service?

What am I doing wrong?

```
"""
Filename: 07_example_timeseries_pandapipes.py
Author: Dipl.-Ing. (FH) Jonas Pfeiffer
Date: 2024-12-06
Description: Script for testing the pandapipes net simulation functions.
Usage: Run the script to generate a simple pandapipes network.
Functions:
[...]

PR #681, opened 2024-12-23
Changed heat consumer flow control when qext_w is zero

JonasPfeiffer123:
In reference to Issue #679 regarding time series calculation problems, I tested setting the mass flow in the heat consumer to zero instead of deactivating the component. This approach seems to resolve the issue I encountered. What are your thoughts on this change? Do you see alternative solutions?

Additionally, I face two further challenges:

- How should the network handle scenarios where all heat consumers have a qext_w of zero in a time step?
- During time steps with minimal overall load, the temperature drops below the heat consumers' return temperature, causing pipeflow to fail due to cooling in the pipes. Increasing text_k in the pipes mitigates the problem, but the results become unrealistic since the ground temperature is not accurately reflected.

What would you recommend to address these challenges?

### DIFF
--- a/src/pandapipes/component_models/heat_consumer_component.py
+++ b/src/pandapipes/component_models/heat_consumer_component.py
@@ -78,12 +78,15 @@ class HeatConsumer(BranchWZeroLengthComponent):
         treturn = net[cls.table_name()].treturn_k.values
         hc_pit[~np.isnan(treturn), TOUTINIT] = treturn[~np.isnan(treturn)]
         hc_pit[:, FLOW_RETURN_CONNECT] = True
-        mask_q0 = qext == 0 & np.isnan(mdot)
+        
+        # Ensure no flow occurs through the heat consumer when qext_w is zero
+        mask_q0 = qext == 0
         if np.any(mask_q0):
-            hc_pit[mask_q0, ACTIVE] = False
+            hc_pit[mask_q0, MDOTINIT] = 0  # Set mass flow to zero
             logger.warning(r'qext_w is equals to zero for heat consumers with index %s. '
-                           r'Therefore, the defined temperature control cannot be maintained.' \
+                           r'Setting mdot to zero to ensure no flow.' \
                     %net[cls.table_name()].index[mask_q0])
+        
         return hc_pit
 
     @classmethod


dlohmeier:
Thanks a lot for the Pull Request @JonasPfeiffer123! I guess that you have already checked the influence on your problem in #679 , can you give some insights to how the temperature control works when setting mdot to 0? The initial idea was that we should handle heat consumers without extracted heat like open valves and set them as "inactive", because then the return temperature control will not be active as well. Of course this can lead to NaN values inside the result table, making it harder to run the next time step. We really need to find a solution for this, but I fear that setting mdot to 0 can have other negative side effects. But I am really curious about your findings.

JonasPfeiffer123:
Thank you for your response, @dlohmeier! I’ve reviewed the influence on my problem and believe I’ve resolved it with this change. When I set mdot to 0 because qext_w is zero, the temperature control does not seem active at the return junction of the heat consumer.

I’ve attached a plot of a test network showing results for a timestep where 2 of the 3 heat consumers have qext_w = 0. As you can see, the return temperature drops at Consumer C's return point, so no temperature is set for A or B. However, when testing other configurations where Consumer C has qext_w = 0 and others are > 0, the pipe flow does not converge. This might be because the pipes/junctions remain active and connected. (Deactivated ends of the network are a problem)

The current setup works because there’s still flow in every pipe, but the second case fails since no flow is present in some pipes. This raises the question: how should the system handle zero flows in pipes? Without flow, temperature cannot be calculated. Would it make sense to assign a default temperature (e.g., 20°C) for junctions with no flow? While it’s not physically accurate, it might suffice given the lack of transient calculations, as the exact temperature in such junctions is not critical. This would also resolve the problem of all heat consumers having qext_w=0. In the timeseries calculation it is also possible to think about using the temperatures from the previous time step.

Minimal Net

dlohmeier:
@SimonRubenDrauz didn't we decide to set the node temperatures to ambient temperature when they are set out of service for heat transfer simulation only? I am not quite sure why this fails here... or is it still part of a feature branch? This should actually solve the problem at hand.

SimonRubenDrauz:
Hi Jonas, thanks a lot for your contribution. We like your solution. We adapted it a bit to make it a bit more stable. See #684. Please check your code with the solution in #684. If you agree we would merge the PR.

@dlohmeier. looking forward for your review as well!

JonasPfeiffer123:
@SimonRubenDrauz  Yeah, I agree.

PR #683, opened 2025-01-06
add documentation for Hofer, add transition area approach

jkisse:
The friction factor "Hofer" is introduced. It is an explicit implementation of Colebrook, used by some industrial Gas Grid Simulation tools.
Please refer to the added paragraph in the [docs](https://pandapipes-jkisse-fork.readthedocs.io/en/feature-hofer/components/pipe/pipe_component.html#friction-models) for more details.

### DIFF
--- a/doc/source/components/pipe/pipe_component.rst
+++ b/doc/source/components/pipe/pipe_component.rst
@@ -156,7 +156,7 @@ They are set by the :code:`friction_model` parameter and the name given in paren
    :nowrap:
 
    \begin{align*}
-    \lambda &= \frac{64}{Re} + \frac{1}{(-2 \cdot \log (\frac{k}{3.71 \cdot d}))^2}\\
+    \lambda &= \frac{64}{Re} + \frac{1}{\left(-2 \cdot \log \left(\frac{k}{3.71 \cdot d}\right)\right)^2}\\
    \end{align*}
 
 
@@ -169,7 +169,7 @@ If *Prandtl-Colebrook* (also known as Colebrook-White) is selected, the friction
    :nowrap:
 
    \begin{align*}
-     \frac{1}{\sqrt{\lambda}} &= -2 \cdot \log (\frac{2.51}{Re \cdot \sqrt{\lambda}} + \frac{k}{3.71 \cdot d})\\
+     \frac{1}{\sqrt{\lambda}} &= -2 \cdot \log \left(\frac{2.51}{Re \cdot \sqrt{\lambda}} + \frac{k}{3.71 \cdot d}\right)\\
    \end{align*}
 
 Equations for pressure losses due to friction were taken from :cite:`Eberhard1990` and
@@ -182,7 +182,7 @@ The *Hofer-Equation* is an explicit approximation of the Prandtl-Colebrook metho
    :nowrap:
 
    \begin{align*}
-    \lambda_H &= \frac{1}{(-2 \log (\frac{4.518}{Re} \cdot \log (\frac{Re}{7}) + \frac{k}{3.71 \cdot d}))^2}\\
+    \lambda_H &= \frac{1}{(-2 \log \left(\frac{4.518}{Re} \cdot \log \left(\frac{Re}{7}) + \frac{k}{3.71 \cdot d}\right)\right)^2}\\
    \end{align*}
 
 Very small Reynolds numbers can lead to problems with the two nested log-functions.
@@ -200,7 +200,7 @@ zone of turbulent flows in pipes and is defined as follows:
    :nowrap:
 
    \begin{align*}
-    \lambda &= \frac{0.25}{(\log(\frac{k}{3.7 \cdot d} + \frac{5.74}{Re^{0.9}}))^2}\\
+    \lambda &= \frac{0.25}{\left(\log\left(\frac{k}{3.7 \cdot d} + \frac{5.74}{Re^{0.9}}\right)\right)^2}\\
    \end{align*}
 
 
--- a/doc/source/components/pipe/pipe_component.rst
+++ b/doc/source/components/pipe/pipe_component.rst
@@ -142,13 +142,15 @@ pipe sections.
 Friction models
 ^^^^^^^^^^^^^^^
 
-Three friction models are used to calculate the velocity dependent friction factor:
+For friction models are available to calculate the velocity dependent friction factor :math:`\lambda`:
 
-- Nikuradse
-- Prandtl-Colebrook
-- Swamee-Jain
+- Nikuradse ("nikuradse")
+- Prandtl-Colebrook ("colebrook")
+- Hofer ("hofer")
+- Swamee-Jain ("swamee-jain")
 
-Nikuradse is chosen by default. In this case, the friction factor is calculated by:
+They are set by the :code:`friction_model` parameter and the name given in parentheses.
+*Nikuradse* is chosen by default. In this case, the friction factor is calculated by:
 
 .. math::
    :nowrap:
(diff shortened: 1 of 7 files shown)


dlohmeier (approved):

jkisse:
Thanks @dlohmeier!
I'd also appreciate your opinion on the usage of np.divide etc. in numba functions, e.g. [here](https://github.com/jkisse/pandapipes/blob/feature/hofer/src/pandapipes/pf/derivative_toolbox_numba.py#L35). It seems to me as if the advantage of array-processing cannot be used in the for-loops, thus it might be slightly more efficient to use the native "/" directly instead of np.divide? Did you investigate that a bit when you implemented the numba-functions?

(analogously for np.abs and np.power)

dlohmeier:
As far as I can recall, numba was good at exactly this: identify numpy code within jit-decorated functions and compile them correctly. I always found it more readable and wrt. timings I remember that I found no difference or a slight advantage of the numpy ufuncs over native code. It might have changed with newer numba versions.

doc/source/components/pipe/pipe_component.rst
```diff
@@ -142,46 +142,65 @@ pipe sections.
 Friction models
 ^^^^^^^^^^^^^^^
 
-Three friction models are used to calculate the velocity dependent friction factor:
+For friction models are available to calculate the velocity dependent friction factor :math:`\lambda`:
```
kbensafta, line 145: small typo in "four"

src/pandapipes/pf/derivative_calculation.py
```diff
@@ -140,10 +140,24 @@ def get_derived_values(node_pit, from_nodes, to_nodes, use_numba):
...
+        lambda_transition[(re >= begin_transition_re) & (re < end_transition_re)]
+    lambda_tot[re >= end_transition_re] = lambda_turb[re >= end_transition_re]
+    return lambda_tot
+
 def calc_lambda(m, eta, d, k, gas_mode, friction_model, lengths, options, area):
     """
     Function calculates the friction factor of a pipe. Turbulence is calculated based on
-    Nikuradse. If v equals 0, a value of 0.001 is used in order to avoid division by zero.
+    Nikuradse. If m equals 0, a value of 0.001 is used in order to avoid division by zero.
     This should not be a problem as the pressure loss term will equal zero (lambda * u^2).
```
kbensafta, line 161: I don't really understand "Turbulence is calculated based on Nikuradse."  It sounds like: Lambda is always calculated using Nikurades in the turbulent interval, regardeless of the choice of the friction model. But in the code, it is indeed the case only when the Nikuardse model is chosen. Am I missing something?

doc/source/components/pipe/pipe_component.rst
```diff
@@ -142,46 +142,65 @@ pipe sections.
...
+   \begin{align*}
+    \lambda_H &= \frac{1}{(-2 \log \left(\frac{4.518}{Re} \cdot \log \left(\frac{Re}{7}) + \frac{k}{3.71 \cdot d}\right)\right)^2}\\
+   \end{align*}
+
+Very small Reynolds numbers can lead to problems with the two nested log-functions.
+Thus, in pandapipes, the laminar :math:`\lambda_l = 64/Re` is applied for small Reynolds numbers
+(:math:`Re < \underline{Re}`) and :math:`\lambda_H` is used for high Reynolds numbers (:math:`Re > \overline{Re}`).
+In the transition area :math:`\underline{Re} \leq Re \leq \overline{Re}`, :math:`\lambda_H` is interpolated linearly.
+By default, the boundaries of this transition area are set to :math:`\underline{Re}=2000` and :math:`\overline{Re}=3000`.
+Note that the derivative used for Hofer is the same as for Prandtl-Colebrook.
```
kbensafta, line 193: Is this indeed the case or is it an approximation, since Hofer is just an explicit from of Prandtl-Colebrook?

src/pandapipes/pf/derivative_calculation.py
jkisse: Probably just some legacy stuff in the docstring. Changed it to point out that the friction_model determines which functions are used.

doc/source/components/pipe/pipe_component.rst
jkisse: fixed

jkisse:
We found out that max_iter has been renamed rather silently a year ago, thus I'd like to use this PR to add a warning for those who still use it.

When calculating large transmissions grids, we often need more than 10 iterations. Therefore, I'd appreciate it if we could set it to 30 by default. It shouldn't have any significant effects on the runtime.

doc/source/components/pipe/pipe_component.rst
jkisse: Not sure - I found it acceptable, also because of the quite complicated derivative of the Hofer formula. Apparently it is working ; )
Maybe the convergence behaviour could be strengthened if a separate "Hofer-derivative" is implemented, but I find it sufficient so far.

dlohmeier:
> We found out that max_iter has been renamed rather silently a year ago, thus I'd like to use this PR to add a warning for those who still use it.
> 
> When calculating large transmissions grids, we often need more than 10 iterations. Therefore, I'd appreciate it if we could set it to 30 by default. It shouldn't have any significant effects on the runtime.

Increasing the default number of iterations is indeed necessary in quite some calculation scenarios for us, so I would welcome that change. Wrt. the max_iter kwarg, I am a little surprised, since I thought we really intended to make this change backwards compatible. A warning is indeed appropriate then.

src/pandapipes/pf/derivative_toolbox.py
```diff
@@ -73,24 +74,45 @@ def derivatives_hydraulic_comp_np(node_pit, branch_pit, lambda_, der_lambda, p_i
...
+    [1]: P. Hofer. Beurteilung von Fehlern in Rohrnetzberechnungen (Error evaluation in calculation
+        of pipelines). GWF–Gas/Erdgas, 114(3):113–119, 1973
+    """
+    lambda_laminar, re = calc_lambda_laminar(m, d, eta, area)
+    lambda_hofer = np.zeros_like(re)
+    re_lower_lim = deepcopy(re)
+    re_lower_lim[re < hofer_re_threshold] = hofer_re_threshold
+    lambda_hofer = \
+        np.divide(1, (-2 * np.log10((4.518/re_lower_lim) * np.log10(re_lower_lim/7) + (k / (3.71 * d)))) ** 2)
+    lambda_hofer[re < hofer_re_threshold] = 0
```
dad616610, line 112: Can be a bit more concise and readable, IMHO

I was getting `invalid value encountered in log10` in the outer log for `re <= 7`, hence the `with np.errstate(...)`. Might be removed, since for all `re < hofer_re_threshold` we're setting `lambda_hofer` to 0, overriding any invalid values

And since `deepcopy` is not used, `from copy import deepcopy` is not needed either
```suggestion
    with np.errstate(invalid="ignore"):
        lambda_hofer = np.where(
            re >= hofer_re_threshold,
            1 / (-2 * np.log10((4.518 / re) * np.log10(re / 7) + (k / (3.71 * d)))) ** 2,
            0,
        )
```

dad616610 (commented):
Some nitpicking

src/pandapipes/pf/derivative_toolbox.py
```diff
@@ -4,6 +4,7 @@
 
 import numpy as np
 from numpy import linalg
+from copy import deepcopy
```
jkisse, line 7: can be removed
```suggestion
```

src/pandapipes/pf/derivative_calculation.py
```diff
@@ -229,9 +249,10 @@ def calc_der_lambda(m, eta, d, k, friction_model, lambda_pipe, area):
     df_dm = np.zeros_like(m)
     df_dlambda = np.zeros_like(m)
     lambda_der = np.zeros_like(m)
-    pos = m != 0
+    pos = ~np.isclose(abs(m), 0)
```
jkisse:
```suggestion
    pos = ~np.isclose(m, 0)
```

jkisse:
> Increasing the default number of iterations is indeed necessary in quite some calculation scenarios for us, so I would welcome that change. Wrt. the max_iter kwarg, I am a little surprised, since I thought we really intended to make this change backwards compatible. A warning is indeed appropriate then.

There is a logger.info if non-standard options are passed to `set_user_pf_options` ([here](https://github.com/e2nIEE/pandapipes/blob/develop/src/pandapipes/pf/pipeflow_setup.py#L203)), but has no effect for options passed as kwargs to pp.pipeflow.

PR #684, opened 2025-01-08
bugfix qzero problem

SimonRubenDrauz:


src/pandapipes/component_models/heat_consumer_component.py
```diff
@@ -170,10 +169,12 @@ def adaption_after_derivatives_hydraulic(cls, net, branch_pit, node_pit, idx_loo
             t_out = hc_pit[:, TOUTINIT]
 
             df_dm = - cp * (t_out - t_in)
-            hc_pit[mask, LOAD_VEC_BRANCHES] = - hc_pit[mask, QEXT] + df_dm[mask] * hc_pit[mask, MDOTINIT]
             mask_equal = t_out == t_in
-            hc_pit[mask & mask_equal, MDOTINIT] = 0
-            hc_pit[mask & ~mask_equal, JAC_DERIV_DM] = df_dm[mask & ~mask_equal]
+            mask_zero = hc_pit[:, QEXT] == 0
+            mask_ign = mask_equal | mask_zero
+            hc_pit[mask & mask_ign, MDOTINIT] = 0
```
dlohmeier, line 175: Is there a case where the derivative could "accidentally" be 0, while mdot is set to 0? I just fear that this could lead to problems with the solver.

src/pandapipes/component_models/heat_consumer_component.py
```diff
@@ -201,9 +202,10 @@ def adaption_after_derivatives_thermal(cls, net, branch_pit, node_pit, idx_looku
         hc_pit = branch_pit[f:t, :]
         consumer_array = get_component_array(net, cls.table_name(), mode='heat_transfer')
 
-        # Any MODE where TRETURN is given
-        mask = consumer_array[:, cls.MODE] == cls.QE_TR
+        mask= consumer_array[:, cls.MODE] == cls.QE_TR
         if np.any(mask):
+            mask_ign = hc_pit[:, QEXT] == 0
+            mask = mask & ~mask_ign
```
dlohmeier, line 208: In this case, the outlet temperature would equal to the inlet temperature, right?

SimonRubenDrauz: Yes!

src/pandapipes/component_models/heat_consumer_component.py
SimonRubenDrauz: No, this would not happen, as the derivative is usually one. Look here:
grafik

SimonRubenDrauz: In all other cases except from the cases where Q and TR are defined JAC_DERIV_DM is equal to one.

dlohmeier (approved):

PR #686, opened 2025-01-16
improving stability and giving hints about possible problems

SimonRubenDrauz:


src/pandapipes/component_models/heat_consumer_component.py
```diff
@@ -76,7 +76,18 @@ def create_pit_branch_entries(cls, net, branch_pit):
         mdot = net[cls.table_name()].controlled_mdot_kg_per_s.values
         hc_pit[~np.isnan(mdot), MDOTINIT] = mdot[~np.isnan(mdot)]
         treturn = net[cls.table_name()].treturn_k.values
-        hc_pit[~np.isnan(treturn), TOUTINIT] = treturn[~np.isnan(treturn)]
+        mask_tr = ~np.isnan(treturn)
+        if np.any(mask_tr):
+            node_pit = net["_pit"]["node"]
+            fn = hc_pit[:, FROM_NODE].astype(int)
+            mask_err = treturn + 15 >= node_pit[fn, TINIT]
```
dlohmeier, line 83: I think that this value of 15K could influence the user to think that a high temperature difference is generally better, although it is purely a matter of the total constellation. Having t_return <= t_in is generally bad, but I would prefer to shift finding a suitable initial temperature difference to a potential diagnostic function.

SimonRubenDrauz: Sounds good to me. Okay, then I will remove this from here.

dlohmeier (approved):

PR #695, opened 2025-03-11
Correct controller documentation

harrahx3:
I found a small mistake in the doc, the function is named run_control, like in pandapower


jkisse (approved):
thanks for your help

PR #700, opened 2025-03-25
Feature/flexible igraph creation

dlohmeier:
extension of build igraph function:

- speedup by using add_edges instead of iterating and avoidance of pandas .loc
- user input for weight column
- user input for additional component models


dlohmeier:
something went wrong here

PR #701, opened 2025-03-25
Feature/ buildup igraph

dlohmeier:
extension of build igraph function
- speedup by using add_edges instead of iterating and avoidance of pandas .loc
- user input for weight column
- user input for additional component models

### DIFF
--- a/src/pandapipes/plotting/generic_geodata.py
+++ b/src/pandapipes/plotting/generic_geodata.py
@@ -19,7 +19,7 @@ logger = logging.getLogger(__name__)
 
 def build_igraph_from_ppipes(net, junctions=None, weight_column_lookup="length_km",
                              edge_factories_override=None, additional_edge_factories=None,
-                             exclude_elements=(), ignore_in_service_elements=()):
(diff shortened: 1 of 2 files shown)


src/pandapipes/plotting/generic_geodata.py
```diff
@@ -44,29 +46,53 @@ def build_igraph_from_ppipes(net, junctions=None):
...
+    if additional_edge_factories is not None:
+        for tbl, edge_factory in additional_edge_factories.items():
+            edge_factory(net, g, pp_junction_mapping, junction_index)
 
     meshed = _igraph_meshed(g)
     roots = [pp_junction_mapping[s] for s in net.ext_grid.junction.values if s in junction_index]
     return g, meshed, roots  # g, (not g.is_dag())
 
 
 def create_generic_coordinates(net, mg=None, library="igraph", geodata_table="junction_geodata",
```
SimonRubenDrauz, line 94: As you allow kwargs, would it make sense, to pass them in functions they make use of this function:
- in create_simple_collections

SimonRubenDrauz (approved):

PR #702, opened 2025-03-26
Fix/json decoder old pandapower

dlohmeier:
For pandapower < 3.0.0, the keyword "ignore_unknown_objects" from PR #665 is not accepted and thus the JSONDecoder raises an error. To avoid this error, this fix checks for the argument in the decoder's and registry's signatures.


dlohmeier:
It appears to me that with pandapower 3.0 and thus the adaptions of pandapipes there are such big changes that it would not make sense to check for backwards compatibility in every regard.

PR #705, opened 2025-04-08
[FIX] fixed imports and dependencies to pandapower 3.0.0

vogt31337:
fixed import if pandapipes is not installed and removed a not found dependency which broke tests


KS-HTK (approved):
Missing the changelog. And better code comments would be appreciated.

PR #712, opened 2025-05-28
Modified output writer to log only existing components to avoid errors

AfaqJ:
This update modifies the init_default_outputwriter function to check whether components exist in the network before logging their variables. Previously, the function unconditionally logged variables for certain components (like sinks, sources, pipes, etc.), which caused errors when those components were missing from the network. By adding checks with hasattr(net, component), the output writer now only logs variables for components that are actually present, improving robustness and preventing runtime errors during time series calculations.


dlohmeier (approved):

dlohmeier:
Looks good to me.

PR #713, opened 2025-06-08
Improve readability of setting pipeflow options

dad616610:
Improve readability and maintainability of `pipeflow_setup`

This PR simplifies the pipeflow initialization logic by:
1. **Making deepcopy operations explicit and singular**
   - Replaces hidden deepcopies in `_iteration_check` and `_check_mode`
   - Eliminates duplicate copying by centralizing the operation
   
2. **Clarifying option precedence**
   - Uses explicit `{**a, **b}` syntax for option merging
   - Clearly defines overwrite order

3. **Removing redundant parameters**
   - Deletes unused default parameters from `pipeflow` function definition
   - (Parameters were always overwritten by `locals()`)

4. **Applying DRY principle**
   - Replaces repetitive branching in `_iteration_check` with loop logic
   - Reduces code duplication across mode checks

Additional improvements:
- Added tests for `_iteration_check` functionality

---

**Planned follow-up work** (separate PRs):
1. Replace `logging.*` calls with `warnings.warn` for:
   - More appropriate user notifications
   - Single-session message suppression
   - Standard Python warning controls
   
2. Add parameter validation in `init_options` to:
   - Surface incorrect values earlier
   - Provide precise error location
   - Improve debugging feedback

### DIFF
--- a/src/pandapipes/pf/pipeflow_setup.py
+++ b/src/pandapipes/pf/pipeflow_setup.py
@@ -279,16 +279,22 @@ def init_options(net, **kwargs):
 
     """
     user_pf_options = net.get("user_pf_options", {})
-    default_opts, user_pf_options, kwargs = map(copy.deepcopy, (default_options, user_pf_options, kwargs))
-    *_, = map(_iteration_check, (user_pf_options, kwargs))
+
+    # prevent mutations
+    default_options_copy = copy.deepcopy(default_options)
+    user_pf_options_copy = copy.deepcopy(user_pf_options)
+    kwargs_copy = copy.deepcopy(kwargs)
+
+    for obj in (user_pf_options_copy, kwargs_copy):
+        _iteration_check(obj)
 
     opts = {
         # Base layer: default options (lowest priority)
-        **default_opts,
-        # Middle layer: network-level options (overrides defaults)
-        **user_pf_options,
+        **default_options_copy,
+        # Middle layer: net options (overrides defaults)
+        **user_pf_options_copy,
         # Top layer: call-specific parameters (highest priority)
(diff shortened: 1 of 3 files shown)


dlohmeier (commented):
I really like all these changes, making the init_option much simpler. I am just wondering about two aspects:
- What does the pipeflow function hand over? It seems to me like a dict, which has to be unpacked in the new structure.
- As we hand over "kwargs" to the init_options function, it might be worth thinking about whether we only allow for kwargs in the pipeflow function and the hand them all over, i.e. make sol_vec a kwarg and add it to the default options.

dad616610:
#### 1) What `pipeflow` hands over to `init_options`
Previously, `pipeflow` passed its entire `locals()` dictionary (containing `net`, `sol_vec`, and `kwargs`) to `init_options`. However, `init_options` only ever used the contents of `kwargs` - the other entries (`net`/`sol_vec`) were unused and immediately discarded.

In this PR, I simplified this by passing only `kwargs` directly:
```python
init_options(**kwargs)
```
This matches the actual requirements of `init_options` while eliminating unnecessary data handling.

#### 2) Handling all parameters via `kwargs`
Currently, `pipeflow` has required parameters (`net`, `sol_vec`) and optional `kwargs`. Since `init_options` doesn't use `net`/`sol_vec` (they're explicitly excluded), passing only `kwargs` is intentional.

If we want certain parameters to be available in `init_options` we could explicitly pass them:
```python
init_options(sol_vec=sol_vec, **kwargs)
```
This would allow `init_options` to access `sol_vec` while keeping other `kwargs`.

dlohmeier (approved):

PR #715, opened 2025-06-16
Feature/valve

SimonRubenDrauz:
enables the direct connection between valve and pipe without a auxiliary junction

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pandapipes"
-version = "0.12.0"  # File format version '__format_version__' is tracked in __init__.py
+version = "0.12.1.dev0"  # File format version '__format_version__' is tracked in __init__.py
 authors=[
     { name = "Simon Ruben Drauz-Mauel", email = "simon.ruben.drauz-mauel@iee.fraunhofer.de" },
     { name = "Daniel Lohmeier", email = "daniel.lohmeier@retoflow.de" },
--- a/src/pandapipes/__init__.py
+++ b/src/pandapipes/__init__.py
@@ -17,7 +17,7 @@ except importlib.metadata.PackageNotFoundError:
             if "project" in data and "version" in data["project"]:
                 __version__ = data["project"]["version"]
 
-__format_version__ = '0.11.1.dev0'
+__format_version__ = '0.12.1.dev0'
 
 import pandas as pd
 import os


src/pandapipes/component_models/component_toolbox.py
```diff
@@ -15,6 +15,18 @@
 from pandapipes.pf.internals_toolbox import _sum_by_group
 
 
+def get_internal_node_lookup_structure(net, internals, table_name, internal_nodes,
+                                       internal_nodes_number, node_start, node_end):
+    internals[table_name] = np.empty((internal_nodes_number, 2), dtype=np.int32)
+    internals[table_name][:, 0] = np.repeat(net[table_name].index, internal_nodes.astype(np.int32))
+    internals[table_name][:, 1] = np.arange(node_start, node_end)
```
dlohmeier, line 22: Should we maybe add another column that gives the internal number of the node on the branch? It could be adjusted by the branch element itself, but is by default just the cumsum for the respective branch index.

src/pandapipes/component_models/pipe_component.py
```diff
@@ -101,31 +115,43 @@ def create_pit_branch_entries(cls, net, branch_pit):
...
+            set_entry_check_repeat(pipe_pit, LENGTH, net[tbl].length_km.values * 1000 / internal_pipe_number,
+                internal_pipe_number, has_internals)
+            set_entry_check_repeat(pipe_pit, K, net[tbl].k_mm.values / 1000, internal_pipe_number, has_internals)
+            set_entry_check_repeat(pipe_pit, ALPHA, net[tbl].u_w_per_m2k.values, internal_pipe_number, has_internals)
+            set_entry_check_repeat(pipe_pit, QEXT, net[tbl].qext_w.values, internal_pipe_number, has_internals)
+            set_entry_check_repeat(pipe_pit, TEXT, net[tbl].text_k.values, internal_pipe_number, has_internals)
+            nan_mask = np.isnan(pipe_pit[:, TEXT])
+            pipe_pit[nan_mask, TEXT] = get_net_option(net, 'ambient_temperature')
+            pipe_pit[:, AREA] = pipe_pit[:, D] ** 2 * np.pi / 4
+            pipe_pit[:, MDOTINIT] *= pipe_pit[:, AREA] * get_fluid(net).get_density(NORMAL_TEMPERATURE)
```
dlohmeier, line 152: I know this is not related to the PR, but I see this for the first time and think that it is prone for errors. Why do we do that?

src/pandapipes/component_models/valve_component.py
```diff
@@ -36,6 +36,64 @@ def active_identifier(cls):
...
+        int_node_pit = super().create_pit_node_entries(net, node_pit)
+        if int_node_pit is not None:
+            if not get_net_option(net, "transient") or get_net_option(net, "simulation_time_step") == 0:
+                junction_table_name = cls.get_connected_node_type().table_name()
+                ft_lookup = get_lookup(net, "node", "from_to")
+                fj_name, _ = cls.from_to_node_cols()
+                f_junction, t_junction = ft_lookup[junction_table_name]
+                junction_pit = node_pit[f_junction:t_junction, :]
+                internal_nodes_lookups = net._lookups['internal_nodes'][cls.table_name()]
+                from_junctions = net[cls.table_name()].loc[internal_nodes_lookups[:, 0], fj_name].values.astype(
```
dlohmeier, line 83: Somehow it seems to me that it would make more sense to have a special lookup for internal nodes. It might be part of a component array, and should hold at least the information if an internal node was created for the respective branch. Then we don't need to do this again here, with info from the internal_nodes_lookup that seems a little tricky with element_idx. This means we need to use .loc instead of an array implementation of the lookup.

src/pandapipes/component_models/valve_component.py
```diff
@@ -36,6 +36,64 @@ def active_identifier(cls):
...
+        :type net: pandapipesNet
+        :return:
+        :rtype:
+        """
+
+        int_nodes = np.zeros(len(net[cls.table_name()]), dtype=np.int32)
+        mask_p = np.flatnonzero(net[cls.table_name()]['et'].values == 'p')
+        val = net[cls.table_name()][['junction', 'element']].iloc[mask_p].values
+        _, idx, inv = np.unique(val, return_index=True, return_inverse=True, axis=0)
+        int_nodes[mask_p[idx]] = 1
```
dlohmeier, line 57: See my comment below: Is this the lookup we need to check on the array-level for e.g. "active" and the "from_junction"?

src/pandapipes/io/convert_format.py
```diff
@@ -57,6 +58,15 @@ def _rename_columns(net):
             for old_col, new_col in list(zip(old_cols, new_cols)):
                 if old_col in net[cp_tbl].columns and new_col not in net[cp_tbl].columns:
                     net[cp_tbl].rename(columns={old_col: new_col}, inplace=True)
+    if Valve.table_name() in net:
+        old_cols = ["from_junction", "to_junction"]
+        new_cols = list(Valve.from_to_node_cols())
+        for o, n in zip(old_cols, new_cols):
+            old_net = False
+            if o in net.valve:
```
dlohmeier, line 66: above, you use Valve.table_name() . I guess it should be kept here.

src/pandapipes/component_models/abstract_models/branch_w_internals_models.py
```diff
@@ -89,29 +86,23 @@ def create_node_lookups(cls, net, ft_lookups, table_lookup, idx_lookups, current
...
         :return:
         :rtype:
         """
-        internal_nodes = cls.get_internal_pipe_number(net) - 1
-        end = current_start
-        ft_lookups[cls.table_name()] = None
+        internal_nodes = cls.get_internal_node_number(net)
+        internal_nodes_num = int(np.sum(internal_nodes))
+        end = current_start + internal_nodes_num
         if np.any(internal_nodes > 0):
```
dlohmeier, line 97: if internal_nodes_num > 0 should do the same and is lighter.

dlohmeier, line 97: if internal_nodes_num > 0 should do the same and is lighter.

src/pandapipes/io/convert_format.py
```diff
@@ -57,6 +58,15 @@ def _rename_columns(net):
...
+                net.valve.rename(columns={o: n}, inplace=True)
+                old_net = True
+            if old_net: net.valve['et'] = 'j'
```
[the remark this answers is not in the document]
dlohmeier: Maybe you could add a check for et that prints a warning that et is overridden.

src/pandapipes/pf/pipeflow_setup.py
```diff
@@ -403,9 +403,9 @@ def initialize_pit(net):
         pit = net["_pit"]
 
     for comp in net['component_list']:
+        comp.create_component_array(net, pit["components"])
```
dlohmeier, line 406: Did you check this? I know that the order was complex and could lead to some issues. I am wondering if we should rather fuse all three methods into one and let the components decide on what to do in which order.

src/pandapipes/component_models/valve_component.py
SimonRubenDrauz: I thought of something similiar as well, but in the end I discarded the idea. The reason: In my eyes a lookup or a component array would only make sense, if you need to call something several times at different locations. However, everything what is done here is done once or are you aware of any duplicated code at this point? 
The real struggle in my eyes is, that you need to combine both internal node/branch lookup of pipes and the internal lookup of valves. This makes it quite complicated. I didn't see any better solution so far. If you have any ideas, I am really open to that.  Long story short: If you agree that evertyhing what is done here needs to be done, cannot be solved more elegantly and is only done once, due to readability I would not introduce a lookup/component array, as it would make everything even more complicated.

src/pandapipes/component_models/valve_component.py
```diff
@@ -47,10 +105,42 @@ def create_pit_branch_entries(cls, net, branch_pit):
...
+                pipe_pit = branch_pit[f:t, :]
+                internal_pipe_lookup = net._lookups['internal_branches']['pipe']
+                internal_nodes_lookups = net._lookups['internal_nodes'][cls.table_name()]
+                mask_p = internal_node_number.astype(bool)
+                pit_idx, lookup_idx = (np.where(
+                    net[cls.table_name()].loc[mask_p, 'element'].values == internal_pipe_lookup[:, 0][:, None]))
+                to_juncts = net['pipe'].loc[net[cls.table_name()].loc[mask_p, 'element'], 'to_junction'].values[
+                    lookup_idx]
+                change_to_nodes = net[cls.table_name()].loc[mask_p, 'junction'].values[lookup_idx] == to_juncts
+                same_pos = lookup_idx[:-1] - lookup_idx[1:]
```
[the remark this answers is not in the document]
SimonRubenDrauz: Is there a more elegant solution in your eyes. I understand that it is complicated but at least everything is at the same location. I have the fear, if we move it to other locations, someone might have struggles to understand how things are connected. Same argument as above: If we need something several times, we should have a lookup. If we need it just once, I would introduce one.

src/pandapipes/pf/pipeflow_setup.py
SimonRubenDrauz: No, however there is no need to do so here. Therefore, I changed it back!

PR #718, opened 2025-06-20
Fix/fix t gen

SimonRubenDrauz:
correction of how to handle input temperature at circ pumps

### DIFF
--- a/src/pandapipes/component_models/abstract_models/circulation_pump.py
+++ b/src/pandapipes/component_models/abstract_models/circulation_pump.py
@@ -6,7 +6,7 @@ import numpy as np
 
 from pandapipes.component_models.abstract_models.branch_wzerolength_models import BranchWZeroLengthComponent
 from pandapipes.component_models.component_toolbox import set_fixed_node_entries, standard_branch_wo_internals_result_lookup
-from pandapipes.idx_branch import D, AREA, LOAD_VEC_BRANCHES_T, TO_NODE, TOUTINIT, JAC_DERIV_DT, JAC_DERIV_DTOUT, QEXT
+from pandapipes.idx_branch import D, AREA, LOAD_VEC_BRANCHES_T, TO_NODE, TOUTINIT, JAC_DERIV_DT, JAC_DERIV_DTOUT, MDOTINIT
 from pandapipes.idx_node import MDOTSLACKINIT, VAR_MASS_SLACK, JAC_DERIV_MSL, NODE_TYPE_T, GE, TINIT
 from pandapipes.pf.pipeflow_setup import get_fluid, get_lookup
 from pandapipes.pf.internals_toolbox import get_from_nodes_corrected
@@ -156,21 +156,31 @@ class CirculationPump(BranchWZeroLengthComponent):
         :type options:
         :return: No Output.
         """
+        node_pit = net['_pit']['node']
+        branch_pit = net['_pit']['branch']
+        branch_lookups = get_lookup(net, "branch", "from_to")
+        f, t = branch_lookups[cls.table_name()]
+
+        mask = branch_pit[f:t, MDOTINIT] < 0
+        if np.any(mask):
+            raise UserWarning(r'Your grid is badly modelled and would lead to a direction change in heat generator {}')
 
         required_results_hyd, required_results_ht = standard_branch_wo_internals_result_lookup(net)
 
         extract_branch_results_without_internals(net, branch_results, required_results_hyd, required_results_ht,
                                                  cls.table_name(), mode)
 
-        node_pit = net['_pit']['node']
-        branch_pit = net['_pit']['branch']
-        branch_lookups = get_lookup(net, "branch", "from_to")
-        f, t = branch_lookups[cls.table_name()]
-
         res_table = net["res_" + cls.table_name()]
 
-        res_table['qext_w'].values[:] = branch_pit[f:t, QEXT]
         from_nodes = get_from_nodes_corrected(branch_pit[f:t])
         t_from = node_pit[from_nodes, TINIT]
         tout = branch_pit[f:t, TOUTINIT]
         res_table['deltat_k'].values[:] = t_from - tout
+
+        fluid = get_fluid(net)
+
+        cp_i = fluid.get_heat_capacity(t_from)
+        cp_i1 = fluid.get_heat_capacity(tout)
+
+        mass = np.abs(branch_pit[f:t, MDOTINIT])
+        res_table['qext_w'].values[:] = mass * (cp_i1 * tout - cp_i * t_from)
--- a/src/pandapipes/component_models/abstract_models/circulation_pump.py
+++ b/src/pandapipes/component_models/abstract_models/circulation_pump.py
@@ -6,9 +6,10 @@ import numpy as np
 
 from pandapipes.component_models.abstract_models.branch_wzerolength_models import BranchWZeroLengthComponent
 from pandapipes.component_models.component_toolbox import set_fixed_node_entries, standard_branch_wo_internals_result_lookup
-from pandapipes.idx_branch import D, AREA, BRANCH_TYPE, CIRC, LOAD_VEC_BRANCHES_T, TO_NODE
-from pandapipes.idx_node import MDOTSLACKINIT, VAR_MASS_SLACK, JAC_DERIV_MSL
-from pandapipes.pf.pipeflow_setup import get_fluid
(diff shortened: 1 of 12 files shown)


src/pandapipes/pf/pipeflow_setup.py
```diff
@@ -561,24 +559,27 @@ def identify_active_nodes_branches(net, hydraulic=True):
...
-            connected_branches_not_zero = branches_connected & ~branches_zero
-            nodes_zero = ~np.isin(
-                np.arange(len(nodes_connected)),
-                np.concatenate([fn[connected_branches_not_zero], tn[connected_branches_not_zero]])
-            ) & nodes_connected
             net["_lookups"]["node_zero_flow"] = nodes_zero
             net["_lookups"]["branch_zero_flow"] = branches_zero
+        else:
+            mask_ge = node_pit[:, NODE_TYPE_T] == GE
+            nodes_connected[mask_ge] = ~nodes_zero[mask_ge]
```
dlohmeier, line 575: is there a connectivity check in every bidirectional calculation? I am wondering if we are producing lots of overhead for this. Wouldn't it make more sense to check for zero flow inside the generator component and use that here?

dlohmeier: I think it would make even more sense to use the definition from before, as stated above in the else-part of the transient check. Any node with zero flow is regarded as disconnected there.

src/pandapipes/component_models/abstract_models/circulation_pump.py
```diff
@@ -152,8 +156,31 @@ def extract_results(cls, net, options, branch_results, mode):
...
+        t_from = node_pit[from_nodes, TINIT]
+        tout = branch_pit[f:t, TOUTINIT]
+        res_table['deltat_k'].values[:] = t_from - tout
+
+        fluid = get_fluid(net)
+
+        cp_i = fluid.get_heat_capacity(t_from)
+        cp_i1 = fluid.get_heat_capacity(tout)
+
+        mass = np.abs(branch_pit[f:t, MDOTINIT])
```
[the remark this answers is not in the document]
SimonRubenDrauz: Just a double check. Chould be removed!

src/pandapipes/component_models/abstract_models/circulation_pump.py
```diff
@@ -100,10 +100,11 @@ def create_pit_branch_entries(cls, net, branch_pit):
         :type branch_pit:
         :return: No Output.
         """
+        circ_pump_tbl = net[cls.table_name()][net[cls.table_name()][cls.active_identifier()].values]
         circ_pump_pit = super().create_pit_branch_entries(net, branch_pit)
         circ_pump_pit[:, D] = 0.1
         circ_pump_pit[:, AREA] = circ_pump_pit[:, D] ** 2 * np.pi / 4
-        circ_pump_pit[:, BRANCH_TYPE] = CIRC
+        circ_pump_pit[:, TOUTINIT] = circ_pump_tbl.t_flow_k.values
```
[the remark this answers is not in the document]
SimonRubenDrauz: Actually, the behaviour remains the same. We just make the pipelfow a bit more stable. Actually I would not add anything or could you think of any case the behaviour would change.

src/pandapipes/pf/pipeflow_setup.py
SimonRubenDrauz: Yes, there is a connectivity check in each bidirectional calculation. Actually, I would say that might be correct, as you always need to check in each heat transfer calculation if the mass flow is unequal zero. 

Therefore, I wouldn't do it in the component itself. I do not see the benefit.

src/pandapipes/component_models/abstract_models/circulation_pump.py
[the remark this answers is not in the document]
dlohmeier: Well, if we fix the outlet temperature instead of the node temperature, I could think of some instances where this does change the result of the pipeflow.

src/pandapipes/pf/pipeflow_setup.py
```diff
@@ -521,64 +519,38 @@ def identify_active_nodes_branches(net, hydraulic=True):
...
-                np.concatenate([fn[connected_branches_not_zero], tn[connected_branches_not_zero]])
-            ) & nodes_connected
-            net["_lookups"]["node_zero_flow"] = nodes_zero
-            net["_lookups"]["branch_zero_flow"] = branches_zero
+            net["_lookups"]["node_zero_flow"] = ~nodes_flow
+            net["_lookups"]["branch_zero_flow"] = ~branches_flow
+        else:
+            branches_connected = branches_flow
+            nodes_connected = nodes_flow
+
```
dlohmeier, line 547: This approach does indeed look much cleaner. But for the zero flow branches / nodes we still need to consider only those active, i.e.

`net["_lookups"]["node_zero_flow"] = ~nodes_flow & nodes_connected`

src/pandapipes/pf/pipeflow_setup.py
```diff
@@ -593,25 +565,12 @@ def branches_connected_flow(branch_pit):
...
-
-    :param branch_pit: The pandapipes internal table of the network (including hydraulics results)
-    :type branch_pit: np.array
-    :return: branches_connected_flow - lookup array if branch is connected wrt. flow
-    :rtype: np.array
-    """
-    # TODO: is this formulation correct or could there be any caveats?
-    return ~np.isnan(branch_pit[:, MDOTINIT]) \
-        & np.isclose(branch_pit[:, MDOTINIT], 0, rtol=1e-10, atol=1e-10)
+    return ~np.isclose(branch_pit[:, MDOTINIT], 0, rtol=1e-10, atol=1e-10)
```
dlohmeier, line 568: Suggestion: rename to branches_not_zero_flow, just to be precise. The connection additionally requires the connected lookup, because nan values would return non-close to zero as well.

src/pandapipes/pf/pipeflow_setup.py
```diff
@@ -645,28 +608,20 @@ def check_connectivity(net, branch_pit, node_pit, mode="hydraulics"):
...
+    return perform_connectivity_search(net, node_pit, branch_pit, slacks,
+                                       nodes_connected, branches_connected, mode=mode)
 
 
 def perform_connectivity_search(net, node_pit, branch_pit, slack_nodes, active_node_lookup, active_branch_lookup,
                                 mode="hydraulics"):
     connect = branch_pit[:, FLOW_RETURN_CONNECT].astype(bool)
-    circ = branch_pit[:, BRANCH_TYPE] == CIRC
-    if np.any(circ) and mode == 'hydraulics':
+    to_nodes = branch_pit[:, TO_NODE]
```
dlohmeier, line 622: Suggestion:

    if mode == "hydraulics":
        connect = branch_pit[:, FLOW_RETURN_CONNECT].astype(bool)
        to_nodes = branch_pit[:, TO_NODE].astype(np.int32)
        gen = node_pit[to_nodes, NODE_TYPE_T] == GE
        if np.any(gen):
            active_branch_lookup = active_branch_lookup & ~connect
    nodes_connected, branches_connected = (
        _connectivity(net, branch_pit, node_pit, active_branch_lookup, active_node_lookup, slack_nodes, mode))
    if np.any(connect) and mode == 'hydraulics':
        from_nodes = branch_pit[:, FROM_NODE].astype(np.int32)
        branch_active = branch_pit[:, ACTIVE].astype(bool)
        active = nodes_connected[from_nodes] & nodes_connected[to_nodes] & branch_active
        branches_connected[connect & active] = True

src/pandapipes/pf/pipeflow_setup.py
```diff
@@ -645,28 +608,20 @@ def check_connectivity(net, branch_pit, node_pit, mode="hydraulics"):
...
+    gen = node_pit[to_nodes.astype(int), NODE_TYPE_T] == GE
+    if np.any(gen) and mode == 'hydraulics':
```
dlohmeier, line 624: Why do we need this gen check here? Is there no other constellation where FLOW_RETURN_CONNECT needs to be considered? What about incomplete networks?

src/pandapipes/pf/pipeflow_setup.py
SimonRubenDrauz: Not required as already done in line 538/539:
```
        nodes_flow = np.isin(np.arange(len(nodes_connected)), np.concatenate([fn[branches_flow], tn[branches_flow]])
        ) & nodes_connected
```

dlohmeier (approved):

PR #726, opened 2025-06-27
Release 0.12 into develop; igraph versioning

EPrade:


pyproject.toml
```diff
@@ -53,7 +53,7 @@ Changelog = "https://github.com/e2nIEE/pandapipes/blob/develop/CHANGELOG.rst"
 
 [project.optional-dependencies]
 docs = ["numpydoc>=1.5.0", "sphinx>=5.3.0", "sphinx_rtd_theme>=1.1.1", "sphinxcontrib.bibtex>=2.5.0", "sphinx-pyproject"]
-plotting = ["plotly", "igraph"]
+plotting = ["plotly", "python-igraph>=0.11.9,<0.12.0"]
```
szhorvat, line 56: Please use `igraph` instead of `python-igraph`. The latter name is deprecated and was intended to be removed some years ago. See https://pypi.org/project/python-igraph/  It will probably stop being supported in the near future.

PR #739, opened 2025-08-04
Improvements STANET converter

dlohmeier:
I implemented an exception for the possibility of an unknown compressibility model in the net parameters of the STANET converter (otherwise the converter fails in an early stage in some cases). The valve pipe creation with individual elements is now vectorized.


SimonRubenDrauz:
At first glance looks good to me. But what about tests. You added new creation modes. Did you test them?

SimonRubenDrauz (approved):

PR #740, opened 2025-08-04
More dynamic fluid properties

dlohmeier:
For compressibility and viscosity, I implemented a possibility to hand over both temperature and pressure as model parameters. This can be important for some very accurate calculations. Any property allowing or requiring to hand over both parameters need the variable "allow_2d" to ensure that the parameterization is correct for each property type.

In addition, two changes were made:
- The Colebrook-White algorithm now uses scipy.optimize.newton root finder instead of hand written code.
- When determining branch intermediate temperatures, we now always use from / to node temperatures in case of gas mode and corrected from / to nodes when checking for branch TOUTINIT (heat transfer calculation).

### DIFF
--- a/.github/workflows/run_tests_develop.yml
+++ b/.github/workflows/run_tests_develop.yml
@@ -94,13 +94,13 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          uv sync --all-extras
+          uv sync --extra all
       - name: List all installed packages
         run: |
           uv pip list
       - name: Test with pytest
         run: |
-          uv sync --extra test --extra plotting
+          uv sync --extra test --extra plotting --extra fileio
           uv run pytest --nbmake -n auto "./tutorials"
       - name: Test without numba
         if: ${{ matrix.python-version == '3.11' }}
--- a/.github/workflows/run_tests_master.yml
+++ b/.github/workflows/run_tests_master.yml
@@ -100,7 +100,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          uv sync --all-extras
+          uv sync --extra all
       - name: List all installed packages
         run: |
           uv pip list


src/pandapipes/pf/derivative_calculation.py
```diff
@@ -359,3 +366,38 @@ def calc_der_lambda(m, eta, d, k, friction_model, lambda_pipe, area):
...
+    :return: lambda_cb
+    :rtype:
+    """
+    def colebrook_white_implicit(lambda_cb, re_nz, k_nz, d_nz):
+        return lambda_cb ** (-1 / 2) + 2 * np.log10(2.51 / (re_nz * np.sqrt(lambda_cb)) + k_nz / (3.71 * d_nz))
+
+    mask = ~np.isclose(re, 0) & ~np.isclose(lengths, 0, rtol=1e-10, atol=1e-11)
+    lambda_res = lambda_nikuradse
+
+    res = newton(colebrook_white_implicit, lambda_res[mask],
```
dad616610, line 393: The current call to `newton` defaults to using secant method (sub-quadratic convergence), because derivative `fprime` of the func is not provided. To achieve faster convergence:
* Newton-Raphson (quadratic convergence): provide derivative `fprime` of the func
* Halley's method (cubic convergence): provide both derivative `fprime` and second order derivative `prime2` of the func

Why? Halley's method converges in fewer steps

See docs for `newton` method: [scipy newton method](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.newton.html)

dlohmeier: I tried your suggestions and remained with the quadratic convergence. That was easy to add, but for my own example, it did not lead to any performance boost. Maybe it helps in some other scenarios. I did struggle with the second derivative, as I got the following problem:

image

Either my derivative has an error (I checked it with sympy and got the same results there, but maybe still there is an error). Gladly have a look at the code I commented out. Or we get problems with very small or very large values here, as we do multiply and divide very large / small numbers. At some point the floating point limits might be reached, to that we get a division by 0 exception. If you have suggestions to get around these problems, I will gladly adapt them into the code. However, I would always prefer a stable approach over a more performant one. It seems to me that scipy is very good at handling such cases internally, so ideally I would hand over as much of the exception checking as possible to scipy or numpy.

dad616610: I agree that stability is paramount. Following your feedback, I've done some research to ensure we're making the best choice, and I'd like to share my findings.

### Halley's Stability
I confirmed that Halley's method is indeed unstable for our problem. Furthermore, even if it were stable, the additional computation of the second derivative makes it slower than the standard Newton method in practice. So, we can conclusively rule it out.

### Performance and Stability of Newton's Method
I extensively tested Newton's method across the full practical range of Reynolds numbers (Re) and relative roughness (ε/D). The key results are:

* Stability: Newton's method converges reliably for the entire tested cartesian product of Re in [2300, 10^16] and ε/D in [0, 0.5].
* Performance: It outperforms all other root-finding methods currently available in scipy.optimize for this specific equation.

This confirms that switching from the secant method to Newton (by providing fprime) is the correct choice for stability and performance, preserving the behavior of our original implementation.

### The Tkachenko-Mileikovskyi Approximation
During testing, I investigated how the quality of the initial guess impacts Newton's performance. This led me to the Tkachenko-Mileikovskyi approximation method. The results were striking:

* Accuracy: As a standalone approximation, its relative error is extremely low. Max relative error for Re [2300, 10^10] and ε/D in [0, 0.5] is 0.00072%. 
<details><summary>Expand to see code confirming the max relative error</summary><p>

```python
import numpy as np
from scipy.optimize import newton

re = np.logspace(3, 10, 100) * 2.3  # [2300, 2.3e10]
ed = np.linspace(0, 0.5, 100) # relative roughness [0, 0.5]

# create cartesian product
re, ed = np.meshgrid(re, ed)
# make 1d (not necessary)
re = re.ravel()
ed = ed.ravel()

def rel_err(exact, approx):
    return (exact - approx) / exact

def tkachenko_mileikovskyi(re_, ed_):
    a0 = -0.79638 * np.log( ed_ / 8.208 + 7.3357 / re_ )
    a1 = re_ * ed_ + 9.3120665 * a0
    return (( 8.128943 + a1 ) / ( 8.128943 * a0 - 0.86859209 * a1 * np.log( a1 / (3.7099535 * re_) ) ))**2

def f(x, re_, ed_):
    return 1/np.sqrt(x) + 2 * np.log10( 2.51 / (re_ * np.sqrt(x)) + ed_ / 3.71 )

ln10 = math.log(10)

def fprime(x, re_, ed_):
    x_32 = x**(-1.5)
    a = 2.51 / re_
    return -0.5 * x_32 - a / (ln10 * ( a * np.sqrt(x) + ed_ / 3.71 ) * x_32 )

tkachenko_mileikovskyi_res = tkachenko_mileikovskyi(re, ed)
newton_res = newton(f=f, fprime=fprime, x0=tkachenko_mileikovskyi_res, args=(re, ed))

r_err = rel_err(newton_res, tkachenko_mileikovskyi_res)
# 0%, 25%, 50%, 75%, 100%
# 0% - min
# 50% - median
# 100% - max
print(np.quantile(r_err, np.linspace(0, 1, 5)))
``` 

</p>
</details> 

* Performance:
        - As a standalone procedure, it's on par with Haaland (~1.5x slower than Swamee-Jain, ~3x slower than Nikuradse).
       - Crucially, when used as the initial guess for Newton, it outperforms all other initial guess methods I tested by a factor of 2x or more. 
<details><summary>Expand to see performance results</summary>
<p>

```
haaland:
        standalone: 0.4175 sec.
        as newton estimate: 6.3397 sec.
swamee_jain:
        standalone: 0.2836 sec.
        as newton estimate: 6.8607 sec.
nikuradse:
        standalone: 0.1320 sec.
        as newton estimate: 9.2353 sec.
tkachenko_mileikovskyi:
        standalone: 0.4466 sec.
        as newton estimate: 3.0581 sec.
```

</p>
</details> 
<details><summary>Expand to see code confirming the performance</summary>
<p>

```python
import numpy as np
from scipy import optimize

def f(x, re, ed):
    return 1/np.sqrt(x) + 2 * np.log10( 2.51 / (re * np.sqrt(x)) + ed / 3.71 )

ln10 = math.log(10)
def fprime(x, re, ed):
    x_32 = x**(-1.5)
    a = 2.51 / re
    return -0.5 * x_32 - a / (ln10 * ( a * np.sqrt(x) + ed / 3.71 ) * x_32 )

def tkachenko_mileikovskyi(re, ed):
    a0 = -0.79638 * np.log( ed / 8.208 + 7.3357 / re )
    a1 = re * ed + 9.3120665 * a0
    return (( 8.128943 + a1 ) / ( 8.128943 * a0 - 0.86859209 * a1 * np.log( a1 / (3.7099535 * re) ) ))**2


def test_speed_approx():
    re = np.logspace(3, 10, 1000) * 2.3
    # relative roughness, 1e-9 because Nikuradse doesn't handle 0
    ed = np.linspace(1e-9, 0.5, 10000)
    re, ed = np.meshgrid(re, ed)
    re = re.flatten()
    ed = ed.flatten()

    meths = {
        "haaland": haaland,
        "swamee_jain": swamee_jain,
        "nikuradse": nikuradse,
        "tkachenko_mileikovskyi": tkachenko_mileikovskyi,
    }
    for name, meth in meths.items():
        t1 = time.perf_counter()
        r = meth(re, ed)
        t2 = time.perf_counter()
        rn = optimize.newton(f, fprime=fprime, x0=r, args=(re, ed))
        t3 = time.perf_counter()
        print(f'{name}:\n\tstandalone: {t2 - t1:.4f} sec.\n\tas newton estimate: {t3 - t2:.4f} sec.')

test_speed_approx()
```

</p>
</details>

This suggests we might not need a full iterative Newton solver at all. We could potentially replace it with the incredibly accurate Tkachenko-Mileikovskyi formula, or at least use it as a lightning-fast initializer for Newton.

### Broader Questions
This research surfaced larger questions about our implementation that are perhaps beyond the scope of this specific PR but deserve their own discussion:

* The Colebrook equation is only physically valid for turbulent flow (Re > ~4000) [wiki](https://en.wikipedia.org/wiki/Darcy_friction_factor_formulae#Colebrook%E2%80%93White_equation). Why do we apply it to the entire range Re > 0?
* The Nikuradse approximation does not handle smooth pipes (ε/D = 0).
* How should we handle the transitional flow regime (2300 < Re < 4000)? Colebrook converges but isn't physically accurate there, while approximation like [Swamee](https://en.wikipedia.org/wiki/Darcy_friction_factor_formulae#Swamee_equation) (not Swamee-Jain) is designed to cover this range.

## Conclusion:
I propose we handle these larger questions in a dedicated new issue and close this thread: we successfully restored the performance of Newton method by providing an `fprime` argument to the function call.

dlohmeier: Thanks a lot for your efforts and sharing your findings. Good to know that the implemented approach is the best wrt. performance and stability. I already thought about using Haaland as an initialization, but if Tkachenko-Mileikovskyi is so much better, it should be worth thinking about this initialization method. We could even offer both of them as options for standalone solving and just let the user choose.

I totally agree that we should have a discussion about the question on how to handle the differences between laminar and turbulent flow regimes + the transition area. When using Nikuradse, I have in mind that this is handled by using a continual approach between the laminar and turbulent equations, which is almost certainly not very precise, but quite common in many applications. Yet, I would prefer to open another issue for that discussion.

PR #769, opened 2025-11-25
Bugfix/plotting valves

EPrade:
Issue: simple_plot for valves took junction geodata and assumed element geodata to also be a junction index. 
This PR adds a check for element type "pi"


kbensafta (approved):
Looks good for me.

PR #773, opened 2025-12-11
New component heat_generator

Theda-ISE:
Adapted heat_consumer derivative calculation in order to operate the heat_consumer with negative power as a secondary heat generator with pump. In a multi producer network it is now possible to use one circ_pressure_pump as a primary heat generator and reverse heat_consumers (defined by a negative power and flow temperature) for the secondary heat generators. The heat_consumers can be modelled with the heat consumers as usual. This replaces the use of circ_pump_mass&Flow control in a multi producer network.

### DIFF
--- a/src/pandapipes/component_models/__init__.py
+++ b/src/pandapipes/component_models/__init__.py
@@ -17,4 +17,5 @@ from pandapipes.component_models.compressor_component import *
 from pandapipes.component_models.flow_control_component import *
 from pandapipes.component_models.mass_storage_component import *
 from pandapipes.component_models.heat_consumer_component import *
+from pandapipes.component_models.heat_generator_component import *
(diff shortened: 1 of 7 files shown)


Theda-ISE:
The recommendation of @SimonRubenDrauz was followed and a new component "heat_generator" was created that merges the functionality of the circulation_pump_pressure (pressure supplying heat generator) and a reverse heat_consumer (heat/mass flow supplying heat generator). Also a new icon was added for the simple_plot function. It should be discussed, if previous change to enable the reverse operation of the heat_consumer with negative power should be maintained as it could be nice to use as a prosumer in certain setups.

EPrade:
I think it would be good to directly include a small tutorial. If you are using the component already, maybe you could add a small jupyter notebook

PR #792, opened 2026-03-16
correction for node_index_active lookup in reduce_pit

dlohmeier:
The node_index_active lookup referred to two different indices depending on the results of the connectivity check. This lead to confusion and certainly wrong behavior, as the lookup could not be properly used in the component models. As this behavior lead to the introduction of the new active_match lookup, this new lookup is hereby removed to reduce the load of information carried along the simulation.


SimonRubenDrauz (approved):

EPrade (approved):

EPrade:
I cannot update this branch with develop, but other than that looks good to me

PR #793, opened 2026-03-23
fix if not all nodes are connected and correction is required

SimonRubenDrauz:


EPrade (approved):

EPrade:
I cannot update this branch with develop, but other than that looks good to me

PR #798, opened 2026-04-15
adding dp losses along a pipe

SimonRubenDrauz:


dlohmeier:
I like the idea of a new metric, but I don't really understand the term dp_loss. From one junction to another, there should be a dp, part of which is due to friction/bending. The other part is due to the height difference and mechanical pressure adaptions (like pumps or throttles). If I understand correctly, by dp_loss you mean exactly the friction/bending part. However, I would call it either p_loss (which might still seem somewhat arbitrary), or dp_friction. Or maybe you have another good idea.

PR #818, opened 2026-07-15
Harden JSON deserialization against arbitrary code execution

kbensafta:
Aligns pandapipes' JSON loading mechanism with the recent security patch in pandapower, [Commit f105e70](https://github.com/e2nIEE/pandapower/commit/f105e70d039bebf95d716783d6f88d52c4f48615#diff-2088af4d80b4ba29a15d15aff8f63978c36ae8121d5219a19b7a6cab947182e7). 
The rest() method in the deserialization registry now uses pandapower's _is_safe_to_deserialize helper to prevent the loading of arbitrary, non-pandapipes-related objects. This closes a remote code execution vulnerability.
Additionally, this commit exposes the skip_checks parameter in from_json and from_json_string to allow trusted files to bypass these new security checks, maintaining API consistency with pandapower.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12"
 ]
 dependencies = [
-    "pandapower == 3.3.3",
+    "pandapower == 3.5.4",
     "matplotlib",
     "shapely",
 ]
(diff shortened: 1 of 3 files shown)


SimonRubenDrauz:
- Please fix the tutorial tests
- check if it works with the most recent pandapower-version. Currently you are using pandapower 3.3.3.

kbensafta:
@SimonRubenDrauz I bumped pandapower from 3.3.3 to 3.5.4 which made the tests pass. But I just noticed that there is PR #817 that addresses the same problem, so I don't know which one we should keep.

PR #820, opened 2026-08-10
Chore/architecture

SimonRubenDrauz:


src/pandapipes/component_models/abstract_models/base_component.py
```diff
@@ -172,28 +99,57 @@ def create_component_array(cls, net, component_pits):
         pass
 
     @classmethod
-    def create_pit_node_entries(cls, net, node_pit):
-        """
-        Function which creates pit branch entries.
+    def register_hydraulic_equations(cls, net, branch_pit, node_pit, sys_idx, registry):
```
dlohmeier, line 102: I am wondering if we could make this approach a little broader and ask the component to register the required equations (or rerun the simulation, see below) with a keyword desceibing the current simulation mode. Then implement the allowed simulation modes for each component. This way it would be possible to add simulation modes
- in a subclass, the superclass call could be used and additional modes implemented in a simple way
- external code could use this component model without having to rewrite too much code

The mode could be based on a string or a Literal.

src/pandapipes/component_models/abstract_models/base_component.py
```diff
@@ -172,28 +99,57 @@ def create_component_array(cls, net, component_pits):
...
+        :param net: a pandapipes net
+        :type net: pandapipes.pandapipesNet
+        :return:
+        :rtype:
         """
-        Function which creates pit branch entries.
+        raise NotImplementedError
+
+    @classmethod
+    def extract_results(cls, net, options, branch_results, mode):
```
dlohmeier, line 142: As mentioned in my above comment, calculation modes would make sense, here we already implemented this approach. We could adapt it once a concept is found for the above methods.

src/pandapipes/pf/system_index.py
```diff
@@ -0,0 +1,397 @@
...
+        if eq.mode == EqWriteMode.UNIQUE:
+            self._check_conflict(eq, self.normal)
+        self.normal.append(eq)
+
+    def add_override(self, eq: ComponentEquations) -> None:
+        if eq.mode == EqWriteMode.UNIQUE:
+            self._check_conflict(eq, self.overrides)
+        self.overrides.append(eq)
+
+    def assemble(self, size: int) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
```
dlohmeier, line 111: I would love to have some more explicit variable names here. It is quite hard to get an overall understanding with such short names, if you need to think about what they mean in the first place.

src/pandapipes/pf/system_index.py
```diff
@@ -0,0 +1,397 @@
...
+        ov_mean     = [e for e in self.overrides if e.mode == EqWriteMode.MEAN]
+
+        def _cat(entries, attr, dtype):
+            return (np.concatenate([getattr(e, attr) for e in entries]).astype(dtype)
+                    if entries else np.empty(0, dtype=dtype))
+
+        n_r  = _cat(n_non_mean, 'rows',      np.int32)
+        n_c  = _cat(n_non_mean, 'cols',      np.int32)
+        n_d  = _cat(n_non_mean, 'data',      np.float64)
+        n_lr = _cat(n_non_mean, 'load_rows', np.int32)
```
dlohmeier, line 143: I have the impression that load vector is a little confusing in this context, even more now that we are naming lines and rows as "load correlated". We could use this opportunity to change the name, e.g. to "residual vector" or "right hand side vector" (which is of course a little long and inconvenient). For the matrix, we might consider to just explicitly state that this is the Jacobian matrix, as system matrix would also make sense for the pits (i.e. a matrix holding all system relevant information).

src/pandapipes/component_models/pipe_component.py
```diff
@@ -128,22 +148,153 @@ def create_pit_branch_entries(cls, net, branch_pit):
...
+        df_dm, df_dp, df_dp1, df_dm_node, load, load_fn, load_tn = (
+            calculate_derivatives_hydraulic(net, branch_pit[f:t], node_pit, get_hydraulic_options(net))
+        )
+
+        b_pit = branch_pit[f:t]
+        fn = b_pit[:, IdxBranch.FROM_NODE].astype(np.int32)
+        tn = b_pit[:, IdxBranch.TO_NODE].astype(np.int32)
+
+        # variables
+        mdot_col   = sys_idx.idx(HydVarEq.MDOTINIT, branch_idx)
```
dlohmeier, line 203: Am I mislead or is this part of the code more or less the same for many branch components? Should we extract it to an external function?

src/pandapipes/component_models/pressure_control_component.py
```diff
@@ -36,110 +45,198 @@ def table_name(cls):
...
+
+        df_dm, df_dp, df_dp1, df_dm_node, load, load_fn, load_tn = (
+            calculate_derivatives_hydraulic(net, b_pit, node_pit, get_hydraulic_options(net))
+        )
+
+        fn = b_pit[:, IdxBranch.FROM_NODE].astype(np.int32)
+        tn = b_pit[:, IdxBranch.TO_NODE].astype(np.int32)
+
+        # Zero out branch equation contributions for ctrl_active branches (replaced by PC constraint)
+        df_dm[ctrl_active]  = 0.0
```
dlohmeier, line 149: Maybe the pressure control component is a good example of using more component-specific registered variables. We don't need mdot if control_active or dp if control is inactive; here just mdot_in = mdot_out, which could be formulated in its own equation. I think, even if we do not adapt the overall structure, we could remove the derivative_calculation call and replace it with very simplified equations, i.e. p_in = p_out for control_inactive and p_out=p_set for control_active.

